PR #9, opened 2020-06-05
APP-2699 - Input component with validation

symphony-adnane:
image

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -1,0 +1,113 @@
+import classNames from 'classnames';
+import { isEmpty } from 'lodash';
+import React from 'react';
+import { ValidatorFn } from '../../core/validators/validators';
+
+type InputProps = {
+  validator?: ValidatorFn | Array<ValidatorFn>;
+  dirty?: boolean;
+  touched?: boolean;
+  label?: string;
+  errors?: { [id: string]: string };
+  onValidationChanged?: (boolean) => any;
+  onChange?: (string) => any;
+  value?: string;
+  placeholder?: string;
+};
+
+export default class Input extends React.Component<InputProps> {
+  public state: any = {
+    dirty: false,
+    touched: false,
+    value: this.props.value || ''
+  };
+  constructor(props) {
+    super(props);
+  }
+
+  get touched() {
+    return this.state.touched;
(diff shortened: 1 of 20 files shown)


src/core/validators/validators.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Factory that returns a validator to test if value matches the provided pattern
+ * Return {pattern:true} if no match,
+ * @param value Value to test
+ */
+const Pattern = (regex: string | RegExp) => {
+  return value => {
+    if (new RegExp(regex).test(value)) {
```
mdeclercq, line 40: If `regex` is not a valid Regex, the component will crash.
That forces the user of the component to put a valid `Regex` pattern and it's a good thing. 

I wonder what's the best approach between this and verifying the Regex before executing it. It would avoid crashes but might hide problems to the user.

src/components/input/Input.tsx
```diff
@@ -0,0 +1,113 @@
...
+      } else {
+        errors = this.props.validator(value);
+      }
+      valid = !errors || isEmpty(errors);
+      if (this.props.onValidationChanged) {
+        this.props.onValidationChanged(valid);
+      }
+    }
+    if (!valid && this.props.errors) {
+      Object.entries(errors).forEach(entry => {
```
mdeclercq, line 65: Instead of `forEach(entry =>`, could you define the variable names of the object key/values, just for clarity ?
Something like `forEach(([validatorName, isInvalid]) =>`.

src/core/validators/validators.ts
symphony-adnane: The validator expects a valid regex indeed
Though a developer should be warned in case of misusing the validator giving it a wrong Regex. 
One option would be to catch, mark as invalid and log an error? 
What do you think?

src/components/input/Input.tsx
symphony-adnane: Absolutely right!!

src/core/validators/validators.ts
mdeclercq: As discussed, let's see what our users think about it. 
Once decided, it should appear clearly in our documentation.

PR #11, opened 2020-06-17
APP-2720 - Support Crop content component

symphony-adnane:
The goal is to support a container component that crops automatically at height (by default at 80px) that could be overriden, it displays a toggle if content is being cropped.
The component reacts to it's size change

## Default
image

## Styled (using tk-quote-container)
image


spec/components/crop-content/CropContent.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+      );
+      wrapper.setProps({ cropHeight: 100 });
+      expect(handleOverflowSpy).toHaveBeenCalled();
+    });
+    it('should handle overflow only on collapsed state', () => {
+      const wrapper = getWrapper();
+      wrapper.setState({ hasOverflow: true });
+      getToggleLink(wrapper).simulate('click');
+
+      const setStateSpy = jest.spyOn(wrapper.instance(), 'setState');
```
micaste, line 74: I'm not a fan of spying on internal methods that much, not on calling methods on the instance directly. If your implementation changes, you'll have to change the tests.
Couldn't you for ex assert that `state.hasOverflow` hasn't changed, rather than checking that setState is not getting called ?

src/core/hoc/ResizeDetectDiv.tsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import * as React from 'react';
+import { withResizeDetector } from 'react-resize-detector';
+
+interface ResizeDetectDivProps {
```
micaste, line 5: I see that you are using types for checking the prop types. I think that this is nice, but it won't raise warnings when the consumer of the library is not using TS, or if there is a layer in-between not using TS. Have you considered that ? Is it much more maintenance to have the types plus the "real" proptypes ?

src/components/crop-content/CropContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+          style={{ maxHeight: height }}
+        >
+          {this.props.children}
+        </div>
+        {/* this should only display if there's overflow or collapsed false */}
+        {this.state.hasOverflow ? (
+          <div className="toggle-container">
+            <a
+              onClick={() => {
+                this.onToggle();
```
micaste, line 68: You might have to bind the function in the constructor, but replacing with
```
onClick={this.onToggle}
```
would save one function creation at each render, plus some listener re-assignation somewhere in react.
A small detail, but I usually do that as a reflex

src/core/hoc/ResizeDetectDiv.tsx
symphony-adnane: I'll do some research on these! thanks for pointing in it out

spec/components/crop-content/CropContent.spec.tsx
symphony-adnane: Agree

src/core/hoc/ResizeDetectDiv.tsx
symphony-adnane: It's worth giving this a try 
https://fettblog.eu/typescript-react/prop-types/#inferring-prop-types

spec/components/crop-content/CropContent.spec.tsx
symphony-adnane: Hi @micaste ,
Actually the test is focusing more on: did we tried to update the hasOverflow? 
Our component's behaviour will do only if it's in collapsed state otherwise it doesn't.
Knowing if the value's changing is kind of impossible since it depends on the containers rendered size (which is not an easy task to do with enzyme, correct me if I'm wrong)

src/components/crop-content/CropContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+  className?: string;
+  collapsed?: boolean;
+  style?: any;
+};
+
+export default class CropContent extends React.Component<CropContentProps> {
+  private containerElRef: HTMLDivElement;
+
+  public state = {
+    collapsed: this.props.collapsed === undefined ? true : this.props.collapsed,
```
micaste, line 16: Ok actually I have a problem with the namings. I see that you have a prop `collapsed` and a state entry named the same. This is a bit confusing for me. It appears that you are only reading the prop when the component is mounted.
I think that I would not expect this, but if it is the behavior you want, you should call the prop `initiallyCollapsed` or something like that.

spec/components/crop-content/CropContent.spec.tsx
micaste: > did we tried to update the hasOverflow?

Why does this matter ? I would only be concerned by the question _"has the `hasOverflow` value changed ?"_
I would agree with your question if you told me _"that's for optimization, to verify if we don't call `setState` too many times"_ to which I would answer  _"react optimizes for you, so you don't really care, and also you're testing too much internals IMO"_
So if you want to answer the question _"has the `hasOverflow` value changed ?"_, you can do `expect(getToggleLink(wrapper)).toBe(undefined)` => actually better than testing the internal state.

> Knowing if the value's changing is kind of impossible since it depends on the containers rendered size

You can do
```
wrapper.instance().containerElRef = {
   offsetHeight: ...,
   scrollHeight: ...,
}
wrapper.find(ResizeDetectDiv).prop("onWidthChange")()
expect(getToggleLink(wrapper)).toBe(undefined)
```

For example.
I'm not hard set on this, but just trying to give you more ideas on what's possible. I usually try to test components "from the outside", not really calling the methods and the states. IMO it makes the tests more robust and needing less updating.
Also, you might want to look into `mount`, that might make the resizing work out of the box, and you might just have to click and assert on node sizes or presence in the dom.

For me the point of unit-tests is to be able to test edge-cases that are otherwise hard to test when using the component in its "functional usage". But if from the outside, without mocking anything, you manage to test all the branches, then I don't see a need to test the internals of calls to the state and things like that.


For example, I see that your method `getWrapper` does not accept props. This means that throughout your unit-test you are not testing your component using different props. Maybe if you did, you wouldn't have to use `wrapper.setState()` etc

symphony-adnane: I don't like testing with internal calls either, thanks for the explanation I'm pretty new to enzyme , need to dig more ^^

src/core/hoc/ResizeDetectDiv.tsx
symphony-adnane: I finally wrote it separately but would definitely go for the inferring for bigger ones

src/components/crop-content/CropContent.tsx
symphony-adnane: Renamed

spec/components/crop-content/CropContent.spec.tsx
symphony-adnane: I used your suggestion to fake size instead of forcing the result through state ;)

PR #174, opened 2021-03-22
 APP-3353 Avatar improvements

symphony-annaLlorens:
## Description
- Exported avatar component as Interface
- Added default avatar on docs
- Bumped UIToolkit Styles

[Jira ticket](https://perzoinc.atlassian.net/browse/APP-3353)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.4.4",
-    "@symphony/uitoolkit-styles": "1.3.6",
+    "@symphony/uitoolkit-styles": "1.3.6-SNAPSHOT.3",
     "@types/classnames": "^2.2.10",
     "classnames": "^2.2.6",
     "date-fns": "^2.16.1",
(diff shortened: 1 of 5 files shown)


symphony-adnane (approved):

stories/Avatar.stories.tsx
```diff
@@ -6,6 +6,19 @@ const img1 = require('./static/avatar.png');
 const img2 = require('./static/avatar1.png');
 const img3 = require('./static/avatar2.png');
 
+const Template = (args) => {
+  return <Avatar {...args}>
+    <img src={img1} alt="avatar" />
+  </Avatar>;
+};
+
+export const Default = Template.bind({});
```
symphony-adnane, line 15: Default is used only for Docs right? 
Just wondering if we could hide it as it's quite misleading when it appears in the list of stories in the left nav...

symphony-annaLlorens: Yes. Maybe we could update all of them without the Default

symphony-jean-michael (approved):

package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.4.4",
-    "@symphony/uitoolkit-styles": "1.3.6",
+    "@symphony/uitoolkit-styles": "1.3.6-SNAPSHOT.3",
```
symphony-adnane, line 32: This looks like a downgrade FYK 
semantically 1.3.6-X.Y < 1.3.6

PR #177, opened 2021-03-22
APP-3922 Security Fix: Set minimum lodash version to 4.17.21

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-3922

**Changes**
- Set minimum lodash version to 4.17.21 because it includes a security fix. (See: https://app.snyk.io/vuln/SNYK-JS-LODASH-1040724)
(FYI, fix by defining the version resolution. See https://classic.yarnpkg.com/en/docs/selective-version-resolutions)

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,9 @@
     "typescript": "^3.9.3",
     "yarn-deduplicate": "^3.1.0"
   },
+  "resolutions": {
+    "react-resize-detector/**/lodash": "^4.17.21"
+  },
   "lint-staged": {
     "src/**/*.{ts,tsx}": [
       "eslint"
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "yarn-deduplicate": "^3.1.0"
   },
   "resolutions": {
-    "react-resize-detector/**/lodash": "^4.17.21"
+    "**/lodash": "^4.17.21"
   },
   "lint-staged": {
     "src/**/*.{ts,tsx}": [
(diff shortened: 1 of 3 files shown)


symphony-annaLlorens (approved):

package.json
```diff
@@ -108,6 +108,9 @@
     "typescript": "^3.9.3",
     "yarn-deduplicate": "^3.1.0"
   },
+  "resolutions": {
+    "react-resize-detector/**/lodash": "^4.17.21"
```
symphony-adnane, line 112: I like this 
we maybe want to do the same for the other libs depending on lodash? 
Did you check if they had a new version not using lodash .20

symphony-jean-michael: You're right. I updated the PR.

package.json
```diff
@@ -20,7 +20,7 @@
     "storybook-docs": "start-storybook --docs",
     "build-storybook": "build-storybook -c .storybook -o dist/storybook",
     "test": "jest --watchAll",
-    "test-once": "jest",
+    "test-once": "jest -w 1",
```
symphony-adnane, line 23: nice why not --maxWorkers=50%?

symphony-adnane (approved):

package.json
symphony-jean-michael: @symphony-adnane I found it in several links. (For example: https://github.com/zcorky/moment/issues/26, https://discuss.circleci.com/t/jest-out-of-memory-error/32263)
So, I propose to try with '-w 1' (This will prevent jest from accidentally spawning too many workers.).
And if the problem still occurs, then we'll try using 50%

PR #185, opened 2021-04-14
APP-4027 - Add right decorator props and deprecate masked prop

AliciaSymphony:
## Ticket
https://perzoinc.atlassian.net/browse/APP-4027

## Description & Demo
- Add new prop `decoratorElement`, the idea is to have the "masked" feature more generic and give the possibility to have something else
- New `isMasked` in order to force the TextField to show "••••••••"
- The `masked` attribute has been deprecated since it's a combination of `decoratorElement` and `isMasked`

image

## Storybook pages
Apr-15-2021 13-04-56
Apr-15-2021 13-05-01

## PR related
Styles https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-styles/pull/80

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -3,6 +3,8 @@ import addons from '@storybook/addons';
 import { addParameters, forceReRender } from '@storybook/react';
 import { themes } from '@storybook/theming';
 
+require('focus-visible');
+
 let globalDarkMode = undefined;
 
 setTimeout(() => init());
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -3,7 +3,7 @@ import addons from '@storybook/addons';
 import { addParameters, forceReRender } from '@storybook/react';
 import { themes } from '@storybook/theming';
 
-require('focus-visible');
+import 'focus-visible';
 
 let globalDarkMode = undefined;
 
--- a/src/components/date-picker/DatePicker.tsx
+++ b/src/components/date-picker/DatePicker.tsx
@@ -484,7 +484,7 @@ class DatePicker extends Component<
             }
             value={inputValue || ''}
             onChange={this.handleInputChange}
-            onFocus={() => this.setState({ showPicker: true })}
+            onClick={() => this.setState({ showPicker: true })}
             onKeyDown={this.handleKeyDownInput}
           ></TextField>
         </div>
--- a/src/components/input/TextComponent.tsx
+++ b/src/components/input/TextComponent.tsx
@@ -139,7 +139,11 @@ const TextComponent: React.FC<TextComponentPropsWithType> = ({
         tooltipCloseLabel={tooltipCloseLabel}
         showRequired={showRequired}
       />
-      <div className="tk-input__container">
+      <div
+        className={classNames(className, 'tk-input__container', {
+          'tk-input__container--disabled': disabled,
+        })}
+      >
         <TagName
           id={id}
           aria-autocomplete="none"
@@ -148,9 +152,7 @@ const TextComponent: React.FC<TextComponentPropsWithType> = ({
           aria-placeholder={placeholder}
           aria-readonly={disabled}
           aria-multiline={type === Types.TEXTAREA}
-          className={classNames('tk-input', className, {
-            'tk-input--with-icon': iconElement,
-          })}
+          className={classNames('tk-input', /**className,**/)}
           placeholder={placeholder}
           value={value}
           onBlur={onBlur}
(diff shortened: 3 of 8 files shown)


.storybook/preview.js
```diff
@@ -3,6 +3,8 @@ import addons from '@storybook/addons';
 import { addParameters, forceReRender } from '@storybook/react';
 import { themes } from '@storybook/theming';
 
+require('focus-visible');
```
symphony-adnane, line 6: why not an import?

src/components/date-picker/DatePicker.tsx
```diff
@@ -480,7 +480,7 @@ class DatePicker extends Component<
             }
             value={inputValue || ''}
             onChange={this.handleInputChange}
-            onFocus={() => this.setState({ showPicker: true })}
+            onClick={() => this.setState({ showPicker: true })}
```
symphony-adnane, line 483: this won't impact keyboard nav?

src/components/input/TextComponent.tsx
```diff
@@ -17,37 +17,47 @@ export type InputBaseProps = {
   onCopy?: (event) => any;
   onCut?: (event) => any;
   onDrag?: (event) => any;
-}
+};
 
 type TextComponentProps = {
+  /** React Element to display inside of the Field, on the right side */
+  decoratorElement?: JSX.Element; // what do you think about the name?
```
symphony-adnane, line 24: I think it should be an array, wouldn't use element as it's redundant 
I'd say
rightDecorators: JSX.Element[]

src/components/input/TextComponent.tsx
```diff
@@ -146,30 +160,37 @@ const TextComponent: React.FC<TextComponentPropsWithType> = ({
...
-        {iconElement && type == Types.TEXTFIELD
-        // Clone the iconElement in order to attach className 'tk-input__icon'
-          ? React.cloneElement(iconElement, {
-            className: classNames(
-              'tk-input__icon',
-              iconElement.props.className
-            ),
-          })
+
+        {decoratorElement && type == Types.TEXTFIELD
```
symphony-adnane, line 172: I suppose that we consider that anything that comes after the input is the right decorator?

src/components/input/TextComponent.tsx
```diff
@@ -146,30 +160,37 @@ const TextComponent: React.FC<TextComponentPropsWithType> = ({
...
+          ? // Clone the decoratorElement in order to attach className 'tk-input__hide'
+            React.cloneElement(decoratorElement)
           : null}
-        {type == Types.TEXTFIELD && masked && value?.length &&  (
+        {type == Types.TEXTFIELD && masked && value?.length ? (
           <button
             className="tk-input__hide"
```
symphony-adnane, line 178: forget to check, but we shouldn't document this class anymore in styles. 
If it's already the case then ignore my comment ;)

src/components/date-picker/DatePicker.tsx
AliciaSymphony: it is actually to prevent to open it with the keyboard ! :D we want only with the click

otherwise with Enter key

src/components/input/TextComponent.tsx
AliciaSymphony: ok let's go for an array

src/components/date-picker/DatePicker.tsx
AliciaSymphony: this change is not related to the fix of the PR, I just embed it here

src/components/input/TextComponent.tsx
AliciaSymphony: We have an example that document this class on storybook styles, I can remove it

stories/TextField.stories.tsx
```diff
@@ -134,9 +137,49 @@ export const TextFields: React.FC = () => {
...
         </p>
-        <TextField value="Lorem" masked={true} onChange={logChange}></TextField>
+        <TextField
+          isMasked={hideText}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          rightDecorators={
+            value?.length
+              ? [
+                <button
```
symphony-adnane, line 150: yeah :)

symphony-adnane (approved):
great @AliciaSymphony

src/components/input/TextComponent.tsx
```diff
@@ -61,13 +71,15 @@ const TextComponentPropTypes = {
   tooltip: PropTypes.string,
   tooltipCloseLabel: PropTypes.string,
   value: PropTypes.string,
-  showRequired: PropTypes.bool
+  showRequired: PropTypes.bool,
 };
 
 const TextComponent: React.FC<TextComponentPropsWithType> = ({
+  rightDecorators,
```
symphony-jean-michael, line 78: A future improvement could be to use a "decorators" (instead of rightDecorators ) and for each decorator define with a prop which side to use (left, right).

symphony-jean-michael (approved):
LGTM

PR #192, opened 2021-04-22
APP-4080 - No sym repo

symphony-adnane:


symphony-jean-michael (approved):

src/core/hoc/Scale.tsx
```diff
@@ -15,6 +15,7 @@ const Scale: React.FC<ScaleProps> = ({
   className,
   ...rest
 }: ScaleProps) => {
+  console.warn('You are using a deprecated component. Scaling won\'t be supported anymore in next major. Please consider removing it');
```
AliciaSymphony, line 18: hooooo!
This is how I can warn when I (we) deprecate a prop for example?

symphony-adnane: that's one way yes :)

symphony-adnane: for props you can also use JSDoc @deprecated jsdoc (could also be used in an ts interface prop for example)

PR #193, opened 2021-04-26
APP-4080 - Go public

symphony-adnane:


symphony-jean-michael (approved):

symphony-annaLlorens (approved):

.circleci/config.yml
```diff
@@ -9,17 +9,15 @@ jobs:
...
       - run: yarn
       - run: yarn lint
       - run: yarn check-circular
       - run:
           command: yarn test-once
           environment:
-              NODE_OPTIONS: --max_old_space_size=4096
+            NODE_OPTIONS: --max_old_space_size=4096
       - run: yarn build
+      - run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
symphony-thibault, line 20: Are we sure that `${NPM_TOKEN}` value does not appear in CircleCI job logs?

symphony-youri: more over do we need a NPM token to access the public registry?

symphony-youri: ah below we public

symphony-adnane: @symphony-thibault CircleCI knows how to handle secrets :)

symphony-jean-michael: @symphony-thibault  Yes, the token will not appear in the logs. the content of the .npmrc file is not displayed. Even if we try, then it will display with '*****' instead of the value. I'll already tried ;)  (For more info: https://circleci.com/docs/2.0/env-vars/#example-configuration-of-environment-variables)

It's seems to be the good way to configure the .npmrc with CircleCI: See https://circleci.com/blog/publishing-npm-packages-using-circleci-2-0/

PR #207, opened 2021-05-24
APP-3937 feat(dropdown): create nested mode

AliciaSymphony:
## Ticket 
https://perzoinc.atlassian.net/browse/APP-3931

## Approach
Create a `mode` prop that can be either nested (or aligned)
- The nested is only applied on the default render
- We don't do anything special in "aligned" mode

Bonus: Reorder the props

## Question
What do you think about the implementation

## Styles PR
https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-styles/pull/90

## Demo
Screenshot 2021-05-24 at 14 31 54

### DIFF
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -29,87 +29,90 @@ import LabelTooltipDecorator from '../label-tooltip-decorator/LabelTooltipDecora
 const prefix = 'tk-select';
 
 export type DropdownProps<T> = {
+  /** Allows to scroll automatically to selected option */
+  autoScrollToCurrent?: boolean;
+  /** Path in custom object to the unique identifier of the option */
+  bindValue?: string;
+  /** Blur the field when an item is selected */
+  blurInputOnSelect?: boolean;
+  /** Optional CSS class name */
+  className?: string;
+  /** Close the expanded menu when the user selects an option */
+  closeMenuOnSelect?: boolean;
+  /** Enables the indicator to expand the Dropdown */
+  displayArrowIndicator?: boolean;
+  /** Display a fixed option on the header of the Dropdown with the searched term */
+  enableTermSearch?: boolean;
+  /** Decides if an item with data and current input value should be displayed in dropdown menu or not */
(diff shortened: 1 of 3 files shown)


src/components/dropdown/Dropdown.tsx
```diff
@@ -326,6 +330,7 @@ export class Dropdown<T = LabelValue> extends React.Component<
           isOptionSelected={this.handleIsOptionSelected}
           menuPlacement="auto"
           maxMenuHeight={maxMenuHeight}
+          mode={mode ? mode : 'aligned'}
```
symphony-annaLlorens, line 333: Why adding the class without any specific css?

symphony-annaLlorens (approved):

src/components/dropdown/Dropdown.tsx
AliciaSymphony: It's used here actually
https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/pull/207/files#diff-19565f42e196c6bb4854c8d839424797c071cb733757393a6d72f1b400ce2f9eR55

I added this line (`mode={mode ? mode : 'aligned'}`) because it was for me the only way to make it available on the custom render (through props?.selectProps?.mode)

Unless there is another way?

src/components/dropdown/Dropdown.tsx
```diff
@@ -29,87 +29,90 @@ import LabelTooltipDecorator from '../label-tooltip-decorator/LabelTooltipDecora
...
-  isInputClearable?: boolean;
-  /** The value of the search input */
-  inputValue?: string;
   /** Maximum height of the menu before scrolling */
   maxMenuHeight?: number;
   /** Max height of the select input before scrolling */
   maxHeight?: number;
-  /** Allows to scroll automatically to selected option */
-  autoScrollToCurrent?: boolean;
+  /** Whether the Dropdown menu is expanded */
```
symphony-adnane, line 73: image

AliciaSymphony: what? XD

symphony-adnane (approved):

PR #213, opened 2021-05-28
APP-3872 Added loading options to Dropdown

symphony-annaLlorens:
### Description 

- Added loading Options dropdown
- Add examples to storybook 

### Demo 

Screenshot 2021-05-28 at 10 27 19

[Jira-ticket](https://perzoinc.atlassian.net/browse/APP-3872)

### DIFF
--- a/src/components/dropdown/CustomRender.tsx
+++ b/src/components/dropdown/CustomRender.tsx
@@ -3,6 +3,7 @@ import classNames from 'classnames';
 import { components } from 'react-select';
 import Icon from '../icon';
 import Loader from '../loader';
+import { SearchHeaderOption } from './interfaces';
 
 /**
  * Useful to stop propagating on mouse down events in custom renderers
@@ -56,38 +57,27 @@ export const DefaultOptionRenderer = (props: any) => {
 // Specific Input to fix input not displayed in React-Select
 // See https://github.com/JedWatson/react-select/issues/3068
 // See https://github.com/JedWatson/react-select/discussions/4302
-export const Input = (props: any) => {
-  const inputAlwaysDisplayed = props?.selectProps?.inputAlwaysDisplayed;
-  return inputAlwaysDisplayed ? 
-    <components.Input {...props} isHidden={false} onKeyUp={props?.selectProps?.onKeyUp}/>
-    : <components.Input {...props} onKeyUp={props?.selectProps?.onKeyUp}/>;
-};
+export const Input = (props: any) => 
+  <components.Input {...props} 
+    isHidden={props?.selectProps?.inputAlwaysDisplayed} 
+    onKeyUp={props?.selectProps?.onKeyUp}
+  />;
 
-export const SingleValue = (props: any) => {
-  const InputRenderer = props.selectProps?.tagRenderer;
-  const inputValue =
-    props.selectProps?.parentInstance?.searchHeaderOption?.value;
-  const rendererProps = { data: props.data };
-  const isSearchHeaderSelected = props?.data?.searchHeader;
-  return (<>
-    {isSearchHeaderSelected ? (
-      <components.SingleValue {...props}>
-        <div>{inputValue}</div>
-      </components.SingleValue>
-    ) : (<>{InputRenderer ? (
-      <components.SingleValue {...props}>
-        <InputRenderer {...rendererProps} />
-      </components.SingleValue>
-    ) : (
-      <components.SingleValue {...props} />
-    )}</>)}
-  </>);
+export const SingleValue =  ({ children, data, selectProps, ...props }: any) => {
+  const InputRenderer = selectProps?.tagRenderer;
+  const inputValue = selectProps?.parentInstance?.searchHeaderOption?.value;
+  const rendererProps = { data };
+  const isSearchHeaderSelected = rendererProps.data.searchHeader;
+  return (
+    <components.SingleValue {...props}>
+      { isSearchHeaderSelected ?  <div>{inputValue}</div> :  (InputRenderer ? <InputRenderer {...rendererProps} /> : children )}
+    </components.SingleValue>);
 };
 
 export const DefaultTagRenderer = (props: any) => {
   const TagRender = props.selectProps?.tagRenderer;
   const rendererProps = { remove: props.removeProps.onClick, data: props.data };
-  return (<div>
+  return (<>
     {TagRender ? 
       <div onMouseDown={stopPropagation}>
         <TagRender {...rendererProps} />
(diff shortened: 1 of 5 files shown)


src/components/dropdown/CustomRender.tsx
```diff
@@ -252,8 +268,12 @@ export const DropdownList = (props: any) => {
     }
     focusThis = focusThis || searchHeaderOption;
     // Clear the value if header option is selected
-    if (select?.state?.selectValue[0]?.searchHeader) {
-      select?.clearValue();
+    // TODO -> Add termSearch to async mode
+    if (select?.state?.selectValue) {
+      if(select?.state?.selectValue[0]?.searchHeader) {
```
AliciaSymphony, line 273: there is a risk of break if select.state.selectValue is empty

src/components/dropdown/Dropdown.tsx
```diff
@@ -277,7 +298,7 @@ export class Dropdown<T = LabelValue> extends React.Component<
           tooltipCloseLabel={tooltipCloseLabel}
           showRequired={showRequired}
         />
-        <Select
+        <Dropdown
```
AliciaSymphony, line 301: I am confused now, from where this is coming from?

src/components/dropdown/CustomRender.tsx
symphony-annaLlorens: done

src/components/dropdown/Dropdown.tsx
symphony-annaLlorens: It will depend if you pass 'asyncOptions' or 'options' as prop

AliciaSymphony (approved):

AliciaSymphony (approved):

src/components/dropdown/CustomRender.tsx
```diff
@@ -35,44 +34,33 @@ export const DefaultOptionRenderer = (props: any) => {
...
-    <components.Input
-      {...props}
-      isHidden={false}
-      onKeyUp={props?.selectProps?.onKeyUp}
-    />
-  ) : (
-    <components.Input {...props} onKeyUp={props?.selectProps?.onKeyUp} />
-  );
+  return inputAlwaysDisplayed ? 
+    <components.Input {...props} isHidden={false} onKeyUp={props?.selectProps?.onKeyUp}/>
```
symphony-adnane, line 62: we're repeating this line just to set isHidden=false? 
why not isHidden={!inputIsAlwaysVisible}

src/components/dropdown/CustomRender.tsx
```diff
@@ -81,168 +69,127 @@ export const SingleValue = (props: any) => {
...
-            </components.SingleValue>
-          ) : (
-            <components.SingleValue {...props} />
-          )}
-        </>
-      )}
-    </>
-  );
+  return (<>
+    {isSearchHeaderSelected ? (
```
symphony-adnane, line 73: this could be refactored to set the child in a var

src/components/dropdown/CustomRender.tsx
```diff
@@ -81,168 +69,127 @@ export const SingleValue = (props: any) => {
...
-export const MultiValueContainerOverride = ({ children, ...props }: any) => {
-  return (
-    <components.MultiValueContainer {...props}>
-      <div>{children}</div>
-    </components.MultiValueContainer>
-  );
+          <div>
+            <Icon
+              iconName="cross"
+              onClick={() => props.removeProps.onClick()}
```
symphony-adnane, line 103: why not props.removeProps.onClick ?

src/components/dropdown/CustomRender.tsx
```diff
@@ -81,168 +69,127 @@ export const SingleValue = (props: any) => {
...
-          ) : (
-            <components.DropdownIndicator {...props} className="tk-d-none" />
-          )}
-        </div>
-      )}
-    </div>
-  );
+  return (<> {isMulti ? 
+    <>{displayArrowIndicator ? 
+      <components.DropdownIndicator {...props}/>
```
symphony-adnane, line 126: too much duplication, we can play instead with classnames and children instead of repeating the whole thing for each condition
it's barely readable

src/components/dropdown/Dropdown.tsx
```diff
@@ -131,11 +130,25 @@ type SingleModeProps<T> = {
   value?: T;
 } & HasValidationProps<T>;
 
+type AsyncProps<T> = {
+  options?: undefined;
+  /** Load the options that populate the dropdown from a returned promise */
+  asyncOptions: (inputValue: string) =>Promise<unknown>;
```
symphony-adnane, line 136: unknown? whouldn't it be T?

src/components/dropdown/Dropdown.tsx
symphony-adnane: sorry i don't get it : where are we using select now? 
thx

src/components/dropdown/interfaces.ts
```diff
@@ -21,4 +21,9 @@ export interface SearchHeaderOption {
   readonly value: string;
 }
 
-export type DropdownOption<T> = T | DropdownOptionsGroup<T>;
\ No newline at end of file
+export type DropdownOption<T> = T | DropdownOptionsGroup<T>;
+
+export const firstOption: Readonly<SearchHeaderOption> = {
```
symphony-adnane, line 26: this isn't really an interface right?

src/components/dropdown/Dropdown.tsx
symphony-adnane: ok got it, does it make sense to store "Dropdown" in the state? 
Also can we use another name because it's very confusing (it seems like we're rendering the same Dropdown component inside Dropdown component...)

src/components/dropdown/CustomRender.tsx
symphony-annaLlorens: done

src/components/dropdown/CustomRender.tsx
symphony-annaLlorens: done

src/components/dropdown/CustomRender.tsx
symphony-annaLlorens: done

src/components/dropdown/CustomRender.tsx
symphony-annaLlorens: done

src/components/dropdown/Dropdown.tsx
symphony-annaLlorens: Its DropdownOption<T>[] 
Done :)

src/components/dropdown/interfaces.ts
symphony-annaLlorens: moved

src/components/dropdown/CustomRender.tsx
symphony-jean-michael: `inputAlwaysDisplayed` is used by TimePicker. I'm not sure the isHidden attribute should be given by default. It may impact the default behavior of Dropdown. I would you suggest to test if there is no regression in this case.

symphony-annaLlorens: Actually by default is false.

symphony-annaLlorens: Updated also

symphony-jean-michael (approved):

PR #224, opened 2021-06-02
Add a wrapper classname for tooltip

liuting1014:
As discussed with @symphony-adnane  and @symphony-jean-michael , we'll expose a wrapper classname for now.  A better solution might come later after further investigation by the team.


src/components/tooltip/Tooltip.tsx
```diff
@@ -71,6 +72,7 @@ const Tooltip: React.FC<TooltipProps> = ({
   placement,
   type,
   visible,
+  wrapperClassName,
```
symphony-adnane, line 75: can't we just name it className? :)

symphony-annaLlorens (approved):

src/components/tooltip/Tooltip.tsx
liuting1014: There is another prop called "className" that will be added to the tooltip itself.

symphony-jean-michael (approved):

src/components/tooltip/Tooltip.tsx
```diff
@@ -106,7 +106,7 @@ const Tooltip: React.FC<TooltipProps> = ({
   });
 
   return (
-    <div ref={ref}>
+    <div ref={ref} className={otherProps.className + '_wrapper'}>
```
symphony-adnane, line 109: maybe if classname is empty we don't want to put anything

symphony-adnane (approved):

PR #228, opened 2021-06-04
APP-4197 Bump UI-Toolkit-Styles v2.0.0

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4197

**Changes**
Bump UI-Toolkit-Styles v2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.4.4",
-    "@symphony-ui/uitoolkit-styles": "1.6.0-rc.1",
+    "@symphony-ui/uitoolkit-styles": "^2.0.0",
     "@types/classnames": "^2.2.10",
     "classnames": "^2.2.6",
     "date-fns": "^2.16.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2244,10 +2244,10 @@
     resolve-from "^5.0.0"
     store2 "^2.12.0"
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.4.4",
-    "@symphony-ui/uitoolkit-styles": "1.6.0-rc.1",
+    "@symphony-ui/uitoolkit-styles": "^2.0.0",
```
AliciaSymphony, line 34: Is "^" on purpose?

AliciaSymphony (approved):

package.json
symphony-jean-michael: Yes, in my point of view, we should no 'lock' on a specific version.

symphony-jean-michael: In our case, we release TK-Components and TK-Styles, so there is no a really impact. But, it's not a good practice to lock on a specific version in the package.json. Because you can prevent those who use the library to get some fixes (released in your dependencies).

Fox example: 
Let 3 libs: A => B => C. (A uses B which uses C)

Imagine we locked the version of C (v2.0.0) in the package.json of B. Next, C provide a security fix (v2.0.1).
Then, A will not be able to get the v2.0.1 of C when it installs B.
Whereas, if the version is not locked, then if 'A' install B, it will get C(v2.0.1)

symphony-annaLlorens (approved):

PR #238, opened 2021-06-15
APP-3518 Async dropdown with different defaultOptions

symphony-annaLlorens:
### Description

With this change, the async dropdown has the possibility to differentiate the options loaded when the dropdown is rendered and when the user starts typing.

### DIFF
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -242,8 +242,8 @@ export class Dropdown<T = LabelValue> extends React.Component<
     }
   }
   
-  internalAsyncOptions = async () => {
-    return this.props?.asyncOptions('')
+  internalAsyncOptions = async (inputValue:string) => {
+    return this.props?.asyncOptions(inputValue)
       .then(options => new Promise(resolve => 
         resolve(this.props.enableTermSearch ?
           [this.searchHeaderOption as T, ...options] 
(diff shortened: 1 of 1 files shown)


symphony-adnane (approved):

src/components/dropdown/Dropdown.tsx
```diff
@@ -136,7 +136,11 @@ type AsyncProps<T> = {
   options?: undefined;
   /** Load the options that populate the dropdown from a returned promise */
   asyncOptions: (inputValue: string) =>Promise<DropdownOption<T>[]>;
-  defaultOptions?: boolean;
+  /**
+   * The default set of options to show before the user starts searching. When
+   * set to `true`, the results for asyncOptions('') will be autoloaded.
+   */
+  defaultOptions?:  DropdownOption<T>[] | boolean;
```
symphony-adnane, line 143: What happens if we provide a list of options? and what happens if we provide a boolean?

symphony-annaLlorens: - If we provide a list of options, they will be loaded when the dropdown is rendered. 
- If we provide a boolean: 
   - > 'true' will load asyncOptions when the dropdown is rendered
   - > 'false' will do nothing

symphony-jean-michael (approved):

AliciaSymphony:
It will be good to have a unit test for this new type of prop

PR #244, opened 2021-06-18
APP-4242 - Tooltip : Wrap the children in a span

AliciaSymphony:
## Ticket
https://perzoinc.atlassian.net/browse/APP-4242

## Description
From bug explained here: 

https://user-images.githubusercontent.com/56824367/122567353-dd107c80-d048-11eb-9a26-cc689d7489f7.mov

https://perzoinc.atlassian.net/wiki/spaces/EIS/pages/1953630703/Weekly+status+meeting+minutes#issue-tooltip-%2F-Switch-disabled
The resolution is to wrap the children inside a span, a solution is to add it directly to the general component, so the workaround doesn't need to be added by the consumer

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -104,13 +104,15 @@ const Tooltip: React.FC<TooltipProps> = ({
       },
     ],
   });
+  
+  const children = <span>{otherProps.children}</span>
 
   return (
     <div ref={ref} className={otherProps.className ? otherProps.className + '_wrapper' : null}>
       <SpanStyled ref={setReferenceElement}>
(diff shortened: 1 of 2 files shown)


symphony-jean-michael (approved):

symphony-annaLlorens (approved):

src/components/tooltip/Tooltip.tsx
```diff
@@ -104,13 +104,15 @@ const Tooltip: React.FC<TooltipProps> = ({
       },
     ],
   });
+  
+  const children = <span>{otherProps.children}</span>
```
symphony-adnane, line 108: why do we need to wrap children in a span?

PR #235, opened 2021-06-11
fix: tooltip behaviour

matthieucrouzet-sym:

PR #246, opened 2021-06-21
fix: tooltip behaviour (#235)

matthieucrouzet-sym:


axeleriksson147:
Would be good if you could provide a description of the PR.

src/components/tooltip/Tooltip.tsx
```diff
@@ -60,8 +62,29 @@ export interface TooltipProps extends Omit<React.HTMLProps<HTMLDivElement>, 'as'
   type?: 'hint' | 'tooltip';
   /** if true, the tooltip should be displayed */
   visible?: boolean;
+  /**
+   * Timeout before the tooltip disappear on hover (in ms)
+   * @default 100
+   */
+  hoverTimeout?: number;
```
axeleriksson147, line 69: If I set displayTrigger to hover, will it automatically disappear after 100ms now?

That would be breaking behaviour with how it works right now.

And if I don't want a hover timeout, how would I set it?

symphony-adnane: @axeleriksson147 how breaking would that be ? 100ms is barely noticeable, but i understand your concern
@matthieucrouzet-sym I understand we need this to allow the user to hover and interact with the tooltip before we close it, do you think we can decrease this value? 
Thanks for your contribution

axeleriksson147: One use case for us is tooltip on hover. You hover over --> tooltip shows until you unhover. With this PR, if I understand it correctly, hovering will disappear after 100ms (which is a very short time to even read what the tooltip says)

matthieucrouzet-sym: Hello @axeleriksson147, in fact this will make the tooltip disappear after 100ms when you are not hovering anymore the tooltip
the goal is to be able to hover the tooltip in order to be able to click on button inside it 
I could set 0ms by default if you want so like that no breaking change and we will create a wrapper component on our side to set the default value to 100ms 
what do you think about that @symphony-adnane  ?

axeleriksson147: Ah, okay! Then it's okay from our point of view. I was confused about the wording of the props:

`* Timeout before the tooltip disappear on hover (in ms)`

PR #257, opened 2021-07-07
feat: add configurable delay to display the Tooltip

nolan-symphony:
We want to add a delay before displaying the Tooltip on Hover

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -67,18 +67,26 @@ export interface TooltipProps
    * @default 100
    */
   hoverTimeout?: number;
+  /**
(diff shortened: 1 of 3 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -213,6 +222,7 @@ Tooltip.propTypes = {
   visible: PropTypes.bool,
   hoverTimeout: PropTypes.number,
   className: PropTypes.string,
+  delayHover: PropTypes.number,
```
AliciaSymphony, line 225: to follow `hoverTimeout` naming, I suggest `hoverDelay`

AliciaSymphony (approved):
Nice improvement!

symphony-adnane (approved):

PR #260, opened 2021-07-15
APP-4318 Fix 'onChange' and 'onBlur' methods are not called in InputDecorator

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4318

**Fix**
Fix 'onChange' and 'onBlur' methods are not called in InputDecorator

**Description of the problem:** The 'onChange' and 'onBlur' were injected by the Validation component and were overwriting those defined on the input.
**Solution:** The inputDecorator calls the onChange and onBlur methods on the Validation, then calls the methods defined on the input.

**Other changes**
- Add jest-extended in devDependencies to be able to use 'toHaveBeenCalledBefore'

### DIFF
--- a/src/components/input/InputDecorator.tsx
+++ b/src/components/input/InputDecorator.tsx
@@ -9,6 +9,7 @@ import {
   LabelTooltipDecoratorPropTypes,
 } from '../label-tooltip-decorator/interfaces';
 import { HasValidationProps } from '../validation/interfaces';
+import { callParentAndChildMethod } from '../../utils';
 
 const RightDecoratorsPropTypes = {
   rightDecorators: PropTypes.oneOfType([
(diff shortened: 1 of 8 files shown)


src/components/input/InputDecorator.tsx
```diff
@@ -73,7 +74,8 @@ const InputDecorator: React.FC<InputDecoratorProps> = ({
     () =>
       child
         ? React.cloneElement(child, {
-          ...rest,
```
AliciaSymphony, line 76: rest is not useful anymore?

AliciaSymphony (approved):

src/components/input/InputDecorator.tsx
symphony-jean-michael: It was a mistake. 'rest' is the args of the InputDecorator.

symphony-annaLlorens (approved):

PR #263, opened 2021-07-19
EIS-3533 - add classname prop to InputDecorator

selena-mallet:
to handle layout we need the prop classname to be added to the InputDecorator Component

Screenshot 2021-07-19 at 15 17 42
Screenshot 2021-07-19 at 14 52 34


symphony-jean-michael (approved):
LGTM. Thank you @selena-mallet

PR #266, opened 2021-07-20
APP-4317 Fix not supported property on Checkbox and fix lint warnings

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4317

**Change:**
Fix not supported property on Checkbox and fix lint warnings

Description: Some props are automatically injected by the Validation Component and are not used in Native HTML tags (like the input). So, I mark them as unused variables in the Component and don't inject them in the native input.

Before the fix:

Screen Shot 2021-07-09 at 11 19 04

With the fix:

Screenshot 2021-07-20 at 11 54 55

### DIFF
--- a/src/components/selection/SelectionInput.tsx
+++ b/src/components/selection/SelectionInput.tsx
@@ -40,6 +40,7 @@ const SelectionInput: React.FC<SelectionInputPropsWithType> = ({
   disabled,
   tabIndex,
   status,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onInit,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onValidationChanged,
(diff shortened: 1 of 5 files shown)


symphony-annaLlorens (approved):

src/components/selection/SelectionInput.tsx
```diff
@@ -39,6 +40,10 @@ const SelectionInput: React.FC<SelectionInputPropsWithType> = ({
   disabled,
   tabIndex,
   status,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
symphony-adnane, line 43: why do we need to extract these if not used?

symphony-adnane (approved):

PR #269, opened 2021-07-21
APP-4191 Select only one option if they have the same value

symphony-annaLlorens:
### Description 

When several options had the same value and the user selects one of them, they were all marked as selected. 

Screenshot 2021-07-21 at 11 14 44

### Fix
When several options had the same value and the user selects one of them, there is only one displayed as selected. 

Screenshot 2021-07-21 at 11 16 25

[Jira ticket](https://perzoinc.atlassian.net/browse/APP-4191)

### DIFF
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -236,7 +236,7 @@ export class Dropdown<T = LabelValue> extends React.Component<
 
   handleIsOptionSelected = this.props.isOptionSelected
     ? (option: any) => this.props.isOptionSelected(option.data)
-    : undefined;
+    :  (option: DropdownOption<T>, selectValue: any) => selectValue?.some(i => i === option);
 
   get internalOptions() {
     if (this.props?.options) {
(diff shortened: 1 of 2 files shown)


src/components/dropdown/Dropdown.tsx
```diff
@@ -236,7 +236,7 @@ export class Dropdown<T = LabelValue> extends React.Component<
 
   handleIsOptionSelected = this.props.isOptionSelected
     ? (option: any) => this.props.isOptionSelected(option.data)
-    : undefined;
+    :  (option: DropdownOption<T>, selectValue: any) => selectValue?.some(i => i === option);
```
symphony-adnane, line 239: could we give a better type to selectValue? it'll be easier to maintain
Also I see you use some on it, are we sure it's always an array or may happen it's a plain object?

symphony-jean-michael (approved):

src/components/dropdown/Dropdown.tsx
symphony-annaLlorens: Type updated. 
I just tested and selectedValue is always an array.

symphony-adnane (approved):

PR #270, opened 2021-07-22
EIS-3537 - Badge default variant

selena-mallet:
- added "default" as a value for the variant prop
- created a BadgeType type
- reorganised files to be consistent with Banner implementation

IMO it is better to have a BadgeType enum instead of hardcoded strings.
It shouldn't be a breaking change as the values of the enum are kept.

the style is now correctly defined:
Screenshot 2021-07-22 at 12 24 00

and we can change on the fly:
Screenshot 2021-07-22 at 12 23 18

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -23,10 +23,10 @@ import Tooltip, { TooltipProps } from './tooltip';
 import Typography from './typography';
 import Validation from './validation';
 import VirtualizedList from './virtualized-list';
-import Badge from './badge';
 
 /* Let's move into exporting everything with interfaces */
 export * from './avatar';
+export * from './badge';
 export * from './banner';
 export * from './button';
 export * from './dropdown';
@@ -60,5 +60,4 @@ export {
   ModalFooter,
   Typography,
   Loader,
-  Badge,
 };
--- a/stories/Badge.stories.tsx
+++ b/stories/Badge.stories.tsx
@@ -11,8 +11,25 @@ export const Default = Template.bind({});
 
 Default.args = {
   children: 'Some Text',
-  variant: ''
 };
+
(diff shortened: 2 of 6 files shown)


symphony-adnane:
Hey Selena this looks good to me 
But we used a typescript type for the variant before
Does this mean that this way we should always go through the Enum to provide a variant? (or can we still write variant="warning")?

stories/Badge.stories.tsx
```diff
@@ -11,38 +11,46 @@ export const Default = Template.bind({});
 
 Default.args = {
   children: 'Some Text',
-  variant: ''
 };
 
+export const Positive: React.FC = () => (
+  <>
+    <Badge variant={BadgeType.POSITIVE} className="tk-mr-h">Badge</Badge>
```
symphony-adnane, line 18: Did we change this because it doesn't work anymore as it was before?

symphony-adnane (approved):

src/components/index.ts
```diff
@@ -23,10 +23,10 @@ import Tooltip, { TooltipProps } from './tooltip';
 import Typography from './typography';
 import Validation from './validation';
 import VirtualizedList from './virtualized-list';
-import Badge from './badge';
 
 /* Let's move into exporting everything with interfaces */
 export * from './avatar';
+export * from './badge';
```
symphony-adnane, line 29: thank you ❤️

AliciaSymphony (approved):
LGTM

stories/Badge.stories.tsx
selena-mallet: it's works as well with the string, you just get an eslint warning.
I thought it was better with an enum, also to stick the Banner component implementation, but I can revert if you don't like.

PR #274, opened 2021-07-28
RTC-10351 - Adding optional className to TextEllipsis

axeleriksson147:
- Added support for className in TextEllipsis
- Added initial support for data-testid to Dropdown component. This needs to be worked on more so all props can be passed correctly but it was a quite tricky component to navigate so I think this is an okay start.

### DIFF
--- a/src/components/text-ellipsis/TextEllipsis.tsx
+++ b/src/components/text-ellipsis/TextEllipsis.tsx
@@ -6,6 +6,9 @@ interface TextEllipsisProps extends Omit<React.HTMLProps<HTMLDivElement>, 'type'
     /** Text that should be ellipsed */
     children?: React.ReactNode;
 
+    /** Optional CSS class name */
+    className?: string;
+  
     /** How many rows the text should span before ellipsing */
     rows?: number;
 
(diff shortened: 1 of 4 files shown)


src/components/text-ellipsis/TextEllipsis.tsx
```diff
@@ -6,6 +6,9 @@ interface TextEllipsisProps extends Omit<React.HTMLProps<HTMLDivElement>, 'type'
     /** Text that should be ellipsed */
     children?: React.ReactNode;
 
+    /** Optional CSS class name */
+    className?: string;
```
symphony-adnane, line 10: thought there would come from the extended props no ?

stories/Dropdown.stories.tsx
```diff
@@ -239,6 +239,7 @@ export const Select: React.FC = () => (
     <h2 className="tk-mt-5h">Scroll into view - disabled</h2>
     <p>{"When pressing the dropdown while it's not fully in view, it will not scroll into view"}</p>
     <Dropdown
+      data-testid="RTC_SETTINGS_DROPDOWN__AUDIO"
```
symphony-adnane, line 242: could we remove this :)

axeleriksson147: Yes, forgot while testing. Fixed!

src/components/text-ellipsis/TextEllipsis.tsx
axeleriksson147: Hm, yes, I've made a mistake.

Removed the prop but kept the part where I put it inside the className={ classnames(...) } property.

symphony-adnane (approved):

AliciaSymphony (approved):

PR #281, opened 2021-08-17
Added a <StylesInjection> HOC that takes care of injecting styles from different styling libraries to the correct document

axeleriksson147:
Created an <UIToolkit> component that takes care of injecting styles from different styling libraries to the correct context. UIToolkit has `styled-components` and  `@emotion` right now that need help injecting their styles to the correct document. Instead of users of this library needing these they need only one import, `<UIToolkit>` with the injectionPoint props that will take of the injection to the correct libraries.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -1,0 +1,15 @@
+# 🧮 Advanced Usage
+
+## 🚏 Using UIToolkit outside the main document
+Some UIToolkit components use styling libraries like `@emotion` and `styled-components` that inject styling into the `<head>` of the main document. If you use these components in a document outside the main one, the styling will only be applied to the main document. You need to use the `<StylesInjection>` component in your root and add an injection point. For example:
+
+```
+import { StylesInjection } from '@symphony-ui/uitoolkit-components';
+
+return (
+    <StylesInjection id="unique_id" injectionPoint={ externalDocument.head }>
+        <Dropdown/>
+        <Tooltip/>
+    </StylesInjection>
+);
+```
--- a/spec/core/hoc/StylesInjection.spec.tsx
+++ b/spec/core/hoc/StylesInjection.spec.tsx
@@ -1,0 +1,13 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { StylesInjection } from '../../../src/core/hoc/StylesInjection'
+
+describe('StylesInjection', () => {
+  
+  it('should render', () => {
+    render(<StylesInjection id="test-id" injectionPoint={document.head}>
+      <p>Hello!</p>
+    </StylesInjection>)
+  })
+
+})
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -25,6 +25,7 @@ import Validation from './validation';
 import VirtualizedList from './virtualized-list';
 
 /* Let's move into exporting everything with interfaces */
+export * from '../core/hoc/index';
 export * from './avatar';
 export * from './badge';
 export * from './banner';
--- a/src/core/hoc/StylesInjection.tsx
+++ b/src/core/hoc/StylesInjection.tsx
@@ -1,0 +1,25 @@
+import * as React from 'react';
+import { StyleSheetManager } from 'styled-components';
+import createCache from '@emotion/cache';
+import { CacheProvider } from '@emotion/react';
+
+interface Props extends React.HTMLProps<HTMLDivElement> {
+    id: string;
+    injectionPoint?: HTMLElement | undefined;
+}
+
+export const StylesInjection = (props: Props) => {
+
+  const emotionCache = createCache({
+    key: props.id,
+    container: props.injectionPoint
+  });
+
+  return (
+    <StyleSheetManager target={ props.injectionPoint }>
+      <CacheProvider value={ emotionCache } >
+        { props.children }
+      </CacheProvider>
+    </StyleSheetManager>
+  )
+}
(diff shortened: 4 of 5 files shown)


symphony-jean-michael (approved):

symphony-adnane:
@axeleriksson147 I feel this is strongly related to how popouts are implemented in Mana 
I'm not sure this documentation (even though very useful) should belong to UIToolkit and even less in the Readme.

axeleriksson147:
@symphony-adnane yes, SFE-Lite does the same thing and we needed to implement similarly in SFE-RTC. I'm fine moving it to some other document but I believe it should be in UIToolkit as it's related to any package wanting to implement it. SFE-Lite and SFE-RTC need to do this cause of how @emotion works which comes from UIToolkit.

Maybe we can change it from popout windows to "external document/iframe". Would that be okay?

docs/styling-engines.md
```diff
@@ -0,0 +1,25 @@
+# 🕴️ Styling engines
```
johanblumenberg-symphony, line 1: Is this documentation intended for the user of UI Toolkit or for the development of UI Toolkit? It seems to be a mix.

docs/styling-engines.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Using UIToolkit in external windows 🪟
+UIToolkit's `Dropdown` and `Tooltip` components have libraries that inject styling with `@emotion`. If you try using these components in a a window outside the main one (popout for example) the styling will only be applied to the main document. So the components needs to be wrapped in a CacheProvider with the container argument set to the document where the styling will be implemented. For example:
```
johanblumenberg-symphony, line 4: Does it work without a `CacheProvider` in the main window? Isn't that only in the special case where the user happens to be an extension deployed in SFE-Lite, where SFE-Lite already added it?
So in fact, every user of UI Toolkit needs to add it?

docs/styling-engines.md
```diff
@@ -0,0 +1,25 @@
...
+return (
+    <CacheProvider value={ emotionCache }>
+        <Dropdown/>
+        <Tooltip/>
+    </CacheProvider>
+);
+```
+
+### Adding styling engines
+If we add a new styling engine, we need to add documentation for how to create an injection point for that specific styling engine.
```
johanblumenberg-symphony, line 23: This would be a breaking change to UI Toolkit

docs/styling-engines.md
```diff
@@ -0,0 +1,25 @@
+# 🕴️ Styling engines
+
+### Using UIToolkit in external windows 🪟
```
johanblumenberg-symphony, line 3: Do you only have to add the CacheProvider around Dropdown and Tooltip?

This would mean that it would be a breaking change if any other component in UI Toolkit starts to use `@emotion`, because everyone that implemented according to this documentation would have to update.

johanblumenberg-symphony: It would be better to have a general statement that all usage of UI Toolkit needs to be wrapped in a CacheProvider. Then we would be safe to continue to use `@emotion ` in all components.

However, if we decide to change to use a different library instead, this would be a breaking change for all users. It would also be a breaking change if we upgrade `@emotion` to a new major version, since all users would have to follow.

Since it is only these two components that use `@emotion`, I think it would be better to completely remove the use of `@emotion` and completely avoid the implications for the user of the library.

johanblumenberg-symphony: cc @symphony-adnane

axeleriksson147: `react-select`, which is the structure of Dropdown uses @emotion, and `Popper` which position the `Tooltip` uses @emotion so it's quite the procedure to get them out of the library. Besides, if we replace it with anything else that uses jss for example @material-ui, we'd need an injection point.

axeleriksson147: Technically you only need it around the `Dropdown` and `Tooltip` but yes I think there should be a general statement for the entire library.

docs/styling-engines.md
axeleriksson147: Yes, it should work without a `CacheProvider` in the main window cause that's where the styling will be injected by default. But anyone using UIToolkit in external windows should wrap that usage in a `CacheProvider` with the new injectionPoint.

docs/styling-engines.md
```diff
@@ -0,0 +1,27 @@
...
+UIToolkit's `Dropdown` and `Tooltip` components have libraries (`popperjs` and `react-select`) that inject styling with `@emotion`. If you try using these components in a a window outside the main one (e.g. popout) the styling will only be applied to the main document.
+
+So if you are using UIToolkit in external windows you need to wrap it in a `CacheProvider` with the container argument set to the document where the styling will be implemented. For example:
+
+```
+import createCache from '@emotion/cache';
+import { CacheProvider } from '@emotion/react';
+
+const emotionCache = createCache({
+    container: popoutDocument.head
```
symphony-jean-michael, line 13: The 'key' is missing in the example (See https://emotion.sh/docs/@emotion/cache)

axeleriksson147: 👍 will add.

src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './components';
+export { UIToolkit } from './styles/UIToolkit'
```
symphony-adnane, line 2: this is still a component, I'd move it under HOC directory in components

symphony-adnane: but let's align first on the solution before applying this

axeleriksson147: Updated after discussion.

symphony-adnane:
do you mind doing codeclimate a favor and adding a test that at least instantiates the component? XD

symphony-adnane (approved):

swasunb (approved):

Issue #280, opened 2021-08-13
[BUG] components/tooltip/Tooltip.tsx


BachLeFPT:
<TooltipContainer
            id={id}
            role="tooltip"
            ref={setPopperElement}
            className={classnames(
              type === 'tooltip' ? 'tk-tooltip' : 'tk-hint',
              **{ className } ==> it always is 'className' in HTML**
            )}
            style={styles.popper}
            {...attributes.popper}
            {...otherProps}
            onMouseEnter={() => handleMouseMove(true)}
            onMouseLeave={() => handleMouseMove(false)}
          >

symphony-adnane:
Hello @BachLeFPT what's the issue here? how could we help ?

BachLeFPT:
@symphony-adnane 
Problem is object in classnames: `{ className }` => always have class `className` in your html, it is not name of class which you pass from props(className)

=> `<div  class=".... className ..." >`

It should be `[className`] instead of `{className}`
(https://github.com/JedWatson/classnames)

```
<TooltipContainer
id={id}
role="tooltip"
ref={setPopperElement}
className={classnames(
type === 'tooltip' ? 'tk-tooltip' : 'tk-hint',
[className]
)}
style={styles.popper}
{...attributes.popper}
{...otherProps}
onMouseEnter={() => handleMouseMove(true)}
onMouseLeave={() => handleMouseMove(false)}
>
```
Thanks!

PR #287, opened 2021-08-30
fix bug 280: classname of tootltip is wrong

tuandva:
## Description
Problem is object in classnames: { className } => always have class className in your html, it is not name of class which you pass from props(className)

It should be [className] instead of {className}
(https://github.com/JedWatson/classnames)

## Issue

[Issue](https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/issues/280)

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -170,7 +170,7 @@ const Tooltip: React.FC<TooltipProps> = ({
             ref={setPopperElement}
             className={classnames(
               type === 'tooltip' ? 'tk-tooltip' : 'tk-hint',
-              { className }
+              className
             )}
             style={styles.popper}
             {...attributes.popper}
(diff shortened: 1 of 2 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -170,7 +170,7 @@ const Tooltip: React.FC<TooltipProps> = ({
             ref={setPopperElement}
             className={classnames(
               type === 'tooltip' ? 'tk-tooltip' : 'tk-hint',
-              { className }
+              [className]
```
symphony-jean-michael, line 173: You can use directly the 'className' instead of using an array. Because it will be ignored if it's a wrong value (undefined, ...). (See https://github.com/JedWatson/classnames)
```
className={classnames(
              type === 'tooltip' ? 'tk-tooltip' : 'tk-hint',
              className
              )}
```

tuandva: thanks. I have changed the code

symphony-adnane (approved):

symphony-jean-michael (approved):

symphony-jean-michael:
ok to test

PR #289, opened 2021-09-07
RTC-10961 - Almost working Tooltip component that doesn't affect child components

axeleriksson147:
Tooltip styling affects the child components. This PR is an attempt to have a working Tooltip without affecting child components.

There's are examples that need to work in TextEllipsis and Tooltip. Right now the only thing not working is <Tooltip> wrapped around an <Icon>. No idea why, it's possible to get it working by adding wrapping elements but that will probably break the example in TextEllipsis. Feedback needed.


src/components/tooltip/Tooltip.tsx
```diff
@@ -142,67 +133,82 @@ const Tooltip: React.FC<TooltipProps> = ({
...
+        ref: setReferenceElement
+      })
+    } else {
+      return <span ref={setReferenceElement}>{child}</span>
+    }
+  })
+
   return (
-    <div ref={ref} className={className ? className + '_wrapper' : null}>
-      <SpanStyled ref={setReferenceElement}>
```
symphony-adnane, line 156: not sure if I understood well this big refactor
But are we removing a span wrapper? 
I think this was mainly added to enable tooltip display on disabled elements (ex a disabled button)

axeleriksson147:
Update: got it working with forwardRef somehow. Will understand better...

src/components/tooltip/Tooltip.tsx
axeleriksson147: Hm, okay.

Seems like it disabled doesn't work with this branch right now.

axeleriksson147: Added back the wrapping element around children props.

src/components/tooltip/Tooltip.tsx
```diff
@@ -48,21 +42,23 @@ const TooltipContainer = styled.div`
...
   /** CSS ID */
   id?: string;
-  /** Function to call on close action */
+  /** Function to call when clicking on closeLabel */
   onHintClose?: () => void;
   placement: 'top' | 'bottom' | 'left' | 'right';
   type?: 'hint' | 'tooltip';
   /** if true, the tooltip should be displayed */
   visible?: boolean;
+  /** Optional CSS class name for wrapping span element  */
```
AliciaSymphony, line 60: I see in the code that we use `div` l.144

AliciaSymphony (approved):
LGTM, just a minor comment

src/components/tooltip/Tooltip.tsx
axeleriksson147: 👍 fixing.

symphony-jean-michael (approved):
LGTM. Small comments/questions but I didn't find a better solution. Thank you so much @axeleriksson147 for your time and your contribution! 😃

src/components/text-ellipsis/TextEllipsis.tsx
```diff
@@ -56,9 +53,10 @@ export const TextEllipsis: React.FC<TextEllipsisProps> = ({
     return(
       <Tooltip
         description={ children as JSX.Element }
-        placement={tooltipPlacement || 'bottom'}
+        placement={ 'top' }
```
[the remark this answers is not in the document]
axeleriksson147: I think "top" is a good default placement for the TextEllipsis. You can override it with the tooltipProps argument.

stories/Tooltip.stories.tsx
```diff
@@ -45,7 +45,7 @@ const addExplanation = (explanation) => (Story) => (
 export const Standard = Template.bind({});
 
 Standard.args = {
-  children: <input type="text" name="fname" />,
+  children: <input type="text" name="fname" onChange={ null } />,
```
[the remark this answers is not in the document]
axeleriksson147: Removing, I had some bug where it was throwing errors in the console cause there was no onChange but not needed.

src/components/tooltip/Tooltip.tsx
```diff
@@ -142,68 +134,79 @@ const Tooltip: React.FC<TooltipProps> = ({
...
-  const children = <span>{otherProps.children}</span>;
-
   const isVisible =
     typeof visible !== 'undefined'
       ? visible
       : displayTrigger === 'hover'
         ? showHover
         : showClick;
 
+  const children = <div
```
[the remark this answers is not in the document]
axeleriksson147: Cause the span height doesn't reflect the child elements so tooltip gets misplaced.

src/components/icon/Icon.tsx
```diff
@@ -33,18 +33,19 @@ const Icon: React.FC<IconProps> = ({
...
     <i
-      {...rest}
+      {...otherProps}
       className={classnames(`tk-icon-${iconName}`, className)}
       onClick={!disabled ? click : null}
       onKeyDown={!disabled ? onKeyDown : null}
       onKeyPress={!disabled ? onKeyPress : null}
       ref={forwardRef}
-      style={{ cursor: !disabled && click && 'pointer' }}
+      style={{ cursor: !disabled && click && 'pointer', ...style }}
```
liuting1014, line 48: Why do we need to pass inline styles? Isn't the classname enough for adding custom stylings?

src/components/button/Button.tsx
```diff
@@ -9,6 +9,7 @@ export type ButtonProps = {
   iconButton?: boolean;
    /** Content of the button*/
   children?: React.ReactNode;
+  /** Optional CSS class name */
```
liuting1014, line 12: It's quite obvious what a classname is so maybe the comment is a bit unnecessary.

src/components/text-ellipsis/TextEllipsis.tsx
```diff
@@ -12,19 +12,16 @@ interface TextEllipsisProps extends Omit<React.HTMLProps<HTMLDivElement>, 'type'
...
 }
 
 export const TextEllipsis: React.FC<TextEllipsisProps> = ({
   children,
   className,
   rows,
   tooltipOnEllipsis,
-  tooltipPlacement,
+  tooltipProps,
+  style,
```
liuting1014, line 24: why inline style again?

src/components/text-ellipsis/TextEllipsis.tsx
liuting1014, line 56: The previous solution looked better because you can clearly see that you can pass in a value for it, or it will default to 'bottom'. This one looks like it's always set to 'top'. Even though you can overwrite it, but just by looking at the code of this component. It's quite unclear.

src/components/tooltip/Tooltip.tsx
```diff
@@ -142,68 +134,79 @@ const Tooltip: React.FC<TooltipProps> = ({
...
+    className={ classnames('tk-tooltip__wrapper', wrapperClassName) }
```
liuting1014, line 145: The `tk-tooltip_wrapper` does not exist in uiToolkit. I previously tried to expose something like `wrapperClassName` but @symphony-adnane mentioned that it's better not to expose it. So we went for a trick like `className={className ? className + '_wrapper' : null}`, this way you won't need to expose a wrapper classname. You can just add your wrapper classname following the format `classname + '_wrapper'` in your `style.ts`.

src/components/tooltip/Tooltip.tsx
```diff
@@ -142,68 +134,79 @@ const Tooltip: React.FC<TooltipProps> = ({
...
+
   return (
-    <div ref={ref} className={className ? className + '_wrapper' : null}>
-      <SpanStyled ref={setReferenceElement}>
-        {displayTrigger === 'hover' &&
+    <>
+      { /**
+       * The element tooltip is wrapped around.
+       */ }
+      { /** Extra span to make sure Tooltip works on disabled children */ }
```
liuting1014, line 156: Where is the span?

src/components/tooltip/Tooltip.tsx
```diff
@@ -212,6 +215,7 @@ Tooltip.defaultProps = {
 };
 
 Tooltip.propTypes = {
+  children: PropTypes.any,
```
liuting1014, line 218: Can we get the correct type instead of `any`?

src/components/icon/Icon.tsx
axeleriksson147: `style` is one of the `React.HTMLProps` and it should work passing inline-styles. className is of course preferred.

src/components/button/Button.tsx
axeleriksson147: Yeah, maybe, just wanted to keep it consistent with other components.

src/components/text-ellipsis/TextEllipsis.tsx
```diff
@@ -12,19 +12,16 @@ interface TextEllipsisProps extends Omit<React.HTMLProps<HTMLDivElement>, 'type'
     /** Wheather a tooltip should be shown on hover when the text is ellipsed */
     tooltipOnEllipsis?: boolean;
 
-    tooltipPlacement?:
-      | 'bottom'
-      | 'left'
-      | 'right'
-      | 'top'
+    tooltipProps?: TooltipProps;
```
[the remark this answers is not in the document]
axeleriksson147: To me it makes sense to be able to pass an entire instance of TooltipProps. We will need wrapperClassName in one place for SFE-RTC and if we start picking out 2-3 props might as well keep the entire one.

src/components/text-ellipsis/TextEllipsis.tsx
axeleriksson147: See previous comment.

src/components/tooltip/Tooltip.tsx
axeleriksson147: Updated to `PropTypes.node` 👍

src/components/tooltip/Tooltip.tsx
axeleriksson147: Moved the comment up to relevant section. Thanks!

src/components/tooltip/Tooltip.tsx
axeleriksson147: I don't think I'm a big fan of the trick. If you need to style the wrapper, you need to know how to do it. It shouldn't be hidden, it's a part of this components interface.

Now it's not a very elegant solution but I think until something better can be thought of we should expose this className wrapper.

src/components/icon/Icon.tsx
axeleriksson147: The reason why it doesn't work is cause we're overriding the style with our own, that's why I have to break it out and pass it in the existing style prop.

src/components/tooltip/Tooltip.tsx
symphony-adnane: wrapper classname to me seems quite an overkill of a prop...
Those are internals of the components IMO we'd be adding a prop for each subcomponent? (which are implementation details that may change in the future...) 
IMHO I'd suggest to use a custom class and append to it __wrapper, and document it 
I let you decide on that

stories/TextEllipsis.stories.tsx
```diff
@@ -59,16 +65,21 @@ export const EllipseAfterTwoRows = (args) => {
...
       >
-        { 'Really, really, really long text that gets cut!' }
+        { 'Icon should be pushed to the right' }
       </TextEllipsis>
+      <Icon iconName="community" style={{ paddingLeft: '8px' }}></Icon>
     </div>
   )
 }
 
+TooltipNotAffectingStyling.decorators = [addExplanation("Use the tooltip prop 'wrapperClassName' to change the styling of the wrapping tooltip element")]
```
symphony-adnane, line 85: :)

src/components/tooltip/Tooltip.tsx
axeleriksson147: After discussion we said it's more clear the implementation with two different classNames. It's not ideal but works for now.

src/components/text-ellipsis/TextEllipsis.tsx
axeleriksson147: Not sure where my comment disappeared or if I never saved it. If okay, I'll leave it like this, didn't wanna break out individual props from Tooltip.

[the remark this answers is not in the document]
AliciaSymphony: Out of scope, but we should at some point add Tooltip examples with bottom, right, left

src/components/button/Button.tsx
AliciaSymphony: I agree that this one is kind of obvious, but as it is reflected in the storybook, it is great to have a short description for every props :D

AliciaSymphony (approved):

PR #292, opened 2021-09-15
RTC-11327 - Let StylesInjection create the id for @emotion/cache

axeleriksson147:
Let StylesInjection create the id for @emotion/cache instead of the one implementing it.

### DIFF
--- a/src/core/hoc/StylesInjection.tsx
+++ b/src/core/hoc/StylesInjection.tsx
@@ -2,16 +2,19 @@ import * as React from 'react';
 import { StyleSheetManager } from 'styled-components';
 import createCache from '@emotion/cache';
 import { CacheProvider } from '@emotion/react';
+import { customAlphabet } from 'nanoid';
 
-interface Props extends React.HTMLProps<HTMLDivElement> {
-    id: string;
+interface Props extends Pick<React.HTMLProps<HTMLDivElement>, 'children'> {
     injectionPoint?: HTMLElement | undefined;
 }
 
 export const StylesInjection = (props: Props) => {
 
+  // @emotion/cache only accept alpha characters
+  const generateUniqueAlphaCharacter = customAlphabet('abcdefghijklmnopqrstuvwxyz', 16);
+
   const emotionCache = createCache({
-    key: props.id,
+    key: generateUniqueAlphaCharacter(),
     container: props.injectionPoint
   });
 
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "emoji-js": "^3.5.0",
     "github-markdown-css": "^4.0.0",
     "lodash": "^4.17.21",
+    "nanoid": "^3.1.25",
     "prop-types": "^15.7.2",
     "react-cool-onclickoutside": "^1.5.8",
     "react-popper": "^2.2.3",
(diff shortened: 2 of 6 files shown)


src/core/hoc/StylesInjection.tsx
```diff
@@ -3,15 +3,17 @@ import { StyleSheetManager } from 'styled-components';
...
-interface Props extends React.HTMLProps<HTMLDivElement> {
-    id: string;
+interface Props extends Pick<React.HTMLProps<HTMLDivElement>, 'children'> {
     injectionPoint?: HTMLElement | undefined;
 }
 
 export const StylesInjection = (props: Props) => {
 
+  // @emotion/cache only accept alpha characters
+  const uniqueAlphaCharacterId = Math.random().toString(36).replace(/[^a-z]+/g, '').substring(0, 10);
```
symphony-jean-michael, line 13: Maybe you could use a lib to generate the ID? In Nexus we use https://github.com/dylang/shortid

src/core/hoc/StylesInjection.tsx
```diff
@@ -3,15 +3,17 @@ import { StyleSheetManager } from 'styled-components';
...
+
   const emotionCache = createCache({
-    key: props.id,
+    key: uniqueAlphaCharacterId,
```
symphony-jean-michael, line 16: IMHO, I would prefer to generate a unique ID if it is not provided in the props. That way there is a way to provide it if needed.

src/core/hoc/StylesInjection.tsx
axeleriksson147: Ok, says it's deprecated so let's use `nanoid`.

However, @emotion/cache only accepts alpha characters so still have to do some conversion.

src/core/hoc/StylesInjection.tsx
axeleriksson147: Resolved, thanks.

src/core/hoc/StylesInjection.tsx
symphony-jean-michael: Thanks for replacing shortId by nanoid :)

src/core/hoc/StylesInjection.tsx
```diff
@@ -2,16 +2,19 @@ import * as React from 'react';
...
 }
 
 export const StylesInjection = (props: Props) => {
 
+  // @emotion/cache only accept alpha characters
+  const generateUniqueAlphaCharacter = customAlphabet('abcdefghijklmnopqrstuvwxyz', 16);
+
   const emotionCache = createCache({
-    key: props.id,
+    key: generateUniqueAlphaCharacter(),
```
symphony-jean-michael, line 17: One of my previous comments seems to have disappeared. 
Could you generate the ID only if it is not provided in props? (To keep a way to specify it if needed)
`key: props.id || generateUniqueAlphaCharacter()`

axeleriksson147: I resolved it so it's hidden.

But I don't think we should allow an id to be passed. There is no point really for users to provide their own id, it will only give them a headache and risk for possible problems if they don't provide a unique id.

symphony-jean-michael: Ok Thanks

symphony-jean-michael (approved):
Thank you @axeleriksson147 :)

AliciaSymphony (approved):
LGTM!

PR #293, opened 2021-09-15
Allow keyboard navigation in dropdowm menu

liuting1014:
Allow user to navigate between menu options using keyboard and activate a certain option by pressing enter. 

https://user-images.githubusercontent.com/32852556/133571267-c9adbb27-d434-48c3-9c7e-079b9e587177.mov

### DIFF
--- a/spec/components/dropdown-menu/DropdownMenu.spec.tsx
+++ b/spec/components/dropdown-menu/DropdownMenu.spec.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {mount} from 'enzyme';
 import {DropdownMenu, DropdownMenuDivider, DropdownMenuItem} from '../../../src/components';
 import {fireEvent, getByText, render, screen} from '@testing-library/react';
+import userEvent, {specialChars} from '@testing-library/user-event';
 
 describe('DropdownMenu', () => {
   it('should render with the correct classes without crash', () => {
@@ -50,5 +51,36 @@ describe('DropdownMenu', () => {
     fireEvent.click(document.body);
     fireEvent.keyUp(document.body, {key: 'Escape', code: 'Escape'});
     expect(close).toHaveBeenCalled();
+  });
+
+  it('should move focus between options by pressing arrow down/up on the current option', () => {
+    render(
+      <DropdownMenu show={true}>
+        <DropdownMenuItem data-testid={'option_1'}>Option 1</DropdownMenuItem>
+        <DropdownMenuItem data-testid={'option_2'}>Option 2</DropdownMenuItem>
+      </DropdownMenu>
+    )
+
+    userEvent.type(screen.getByTestId('option_1'), specialChars.arrowDown);
+    expect(screen.getByTestId('option_2')).toHaveFocus();
+
+    userEvent.type(screen.getByTestId('option_2'), specialChars.arrowUp);
+    expect(screen.getByTestId('option_1')).toHaveFocus();
+
+    // reaches the end
+    userEvent.type(screen.getByTestId('option_1'), specialChars.arrowUp);
+    expect(screen.getByTestId('option_2')).toHaveFocus();
+  })
+
+  it('should call onClick when pressing enter on an option', () => {
+    const onClick = jest.fn();
+    render(
+      <DropdownMenu show={true}>
+        <DropdownMenuItem data-testid={'option_1'} onClick={onClick}>Option 1</DropdownMenuItem>
+      </DropdownMenu>
+    )
+
+    userEvent.type(screen.getByTestId('option_1'), specialChars.enter);
+    expect(onClick).toHaveBeenCalled();
   })
 })
--- a/src/components/dropdown-menu/DropdownMenu.tsx
+++ b/src/components/dropdown-menu/DropdownMenu.tsx
@@ -13,19 +13,47 @@ interface DropdownMenuProps extends React.HTMLProps<HTMLDivElement> {
 interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   className?: string;
-  onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+  /** To select a certain option in the menu without having to use document.querySelector */
+  forwardRef?: React.RefObject<HTMLDivElement>;
+  onClick?: (event: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>) => void;
 }
 
 export const DropdownMenuDivider: React.FC = () => <div className="tk-dropdown-menu-divider"></div>
 
-export const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({children, className, onClick, ...rest}: DropdownMenuItemProps) => {
+export const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({children, className, onClick, forwardRef, ...rest}: DropdownMenuItemProps) => {
   const classes = classNames(
     'tk-dropdown-menu__item',
     className,
   )
 
+  const focusNextOption = (current: HTMLDivElement, direction: number) => {
+    const options = (Array.from(current.parentElement.getElementsByClassName('tk-dropdown-menu__item'))) as HTMLDivElement[];
+    let currentOptionPosition = options.indexOf(current);
+    currentOptionPosition = currentOptionPosition > -1 ? currentOptionPosition : 0;
+    const nextElementPosition = (currentOptionPosition + options.length + direction) % options.length;
+
+    options[nextElementPosition].focus();
+  }
+
+  const onKeyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    switch (e.key) {
+    case Keys.ARROW_DOWN :
+      e.stopPropagation();
+      focusNextOption(e.currentTarget, 1);
+      break;
+    case Keys.ARROW_UP:
+      e.stopPropagation();
+      focusNextOption(e.currentTarget, -1);
+      break;
+    case Keys.ENTER:
+      e.stopPropagation();
+      onClick(e);
+      break;
+    }
+  }
+
   return (
-    <div {...rest} className={classes} onClick={onClick}>
+    <div {...rest} className={classes} onClick={onClick} ref={forwardRef} onKeyDown={onKeyDownHandler} tabIndex={-1}>
       {children}
     </div>
   )
@@ -66,5 +94,5 @@ export const DropdownMenu: React.FC<DropdownMenuProps> = ({children, className, 
       {children}
     </div>
   )
-}
+};
 


axeleriksson147 (approved):
Nice!

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   className?: string;
   onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+  // To select a certain option in the menu without having to use document.querySelector
+  customRef?: React.RefObject<HTMLDivElement>;
+  onActivate?: (event: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>) => void;
```
johanblumenberg-symphony, line 19: Perhaps remove the `onClick` event, and only use `onActivate`? Or perhaps only keep `onClick` and call that one if the user activates the button as well? How does a regular `<div>` tag handle clicks and pressing enter? We should do the same.

How does a regular `<button>` or `<div>` work? I think we should be really close to how native HTML works, and do exactly the same. Or at least the same way that `react` handles these elements. I believe they just call the `onClick` event when you press enter.

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+  const onKeyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    e.stopPropagation();
+    switch (e.key) {
+    case 'ArrowDown' :
+      focusNextOption(e.currentTarget, 1);
+      break;
+    case 'ArrowUp':
+      focusNextOption(e.currentTarget, -1);
+      break;
+    case 'Enter':
```
johanblumenberg-symphony, line 48: I'm not sure you need to handle the `Enter` key.

If you don't add any code at all, the browser will handle pressing enter. So it is strange if you need to add code for this case.

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
 
-export const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({children, className, onClick, ...rest}: DropdownMenuItemProps) => {
+export const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({children, className, onClick, customRef, onActivate, ...rest}: DropdownMenuItemProps) => {
   const classes = classNames(
     'tk-dropdown-menu__item',
     className,
   )
 
+  const focusNextOption = (current: HTMLDivElement, direction: number) => {
+    const options = (Array.from(current.parentElement.getElementsByClassName('tk-dropdown-menu__item'))) as HTMLDivElement[];
```
johanblumenberg-symphony, line 31: Is it possible to use `current.nextElementSibling` or `current.prevElementSibling`?

Those will give you exactly the elements you are looking for, right?

Selecting based on classes means you select based on the styling. If the way the element is styled changes, the navigation also breaks. Adding wrapper elements, or adding child elements with the same class will also confuse the navigation code. Selecting by class name gives you what you need right now, but there are many future code changes that are completely unrelated to keyboard navigation that would break your code.

I think it would be better to write the navigation code completely independent from styling.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: I think it's better to keep the onClick for backward compatibility.  I will change it to call the onClick when user activates the option as well.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: The problem with doing `current.nextElementSibling` or `current.prevElementSibling` is that when you reach the beginning or end of the list, it won't jump to the first/last option as there is no previous or next sibling anymore, which is not the desirable experience. You would expect it to focus the first option when you press arrowDown on the last option, and the other way around.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: As far as I know, I think it only works on elements like form, button, checkbox... but not a div.

src/components/dropdown-menu/DropdownMenu.tsx
axeleriksson147: Could you use `current.nextElementSibling` and `current.previousElementSibling`, and if they are undefined use the `current.parent.lastChild` and `current.parent.firstChild`. Shouldn't that work?

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: Actually there is a problem with calling the onClick when pressing enter, one is a keyboard event and one is a mouse event. I would have to forcefully cast it to any to be able to call onClick. Maybe it's fine to keep both.

axeleriksson147: `React.HTMLProps<HTMLDivElement>` doesn't it contain all these props already? `onClick`, 'className` and `onClose` so them being defined is unnecessary.

But what's the difference between `onActivate` and `onClick`? Why do we need both? Shouldn't they do the same.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: You can have other non-menu options as well, such as a divider. and you don't want to focus those.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: > `React.HTMLProps<HTMLDivElement>` doesn't it contain all these props already? `onClick`, 'className`and`onClose` so them being defined is unnecessary.
> 
Yes, I can remove those. The React.HTMLProps<HTMLDivElement> was added later, that's why those were there before.

> But what's the difference between `onActivate` and `onClick`? Why do we need both? Shouldn't they do the same.
The onActivate is when you activate the menu option, doesn't matter if it's through keyboard or mouse.

axeleriksson147: 👍 thanks!

src/components/dropdown-menu/DropdownMenu.tsx
axeleriksson147: Okay, tricky..

src/components/dropdown-menu/DropdownMenu.tsx
johanblumenberg-symphony: https://stackoverflow.com/questions/32659099/making-a-clickable-div-accessible-through-tab-structure/32659785

The accepted answer here makes some good points about creating accessible elements. Perhaps it would be better to use a `<button>` element, not sure...

I don't have enough knowledge about how to make an app accessible.

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
   children?: React.ReactNode;
   className?: string;
   onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+  // To select a certain option in the menu without having to use document.querySelector
```
AliciaSymphony, line 17: Can you document the props using `/** My description */` thank you :)

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+  customRef?: React.RefObject<HTMLDivElement>;
```
AliciaSymphony, line 18: Minor detail, in the icon component we name it `forwardRef`, I think that you have the same use here, do you mind aligning the name please?
https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/blob/master/src/components/icon/Icon.tsx#L12

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: That would've been easier, but using button will break the stylings though...

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+    let currentOptionPosition = options.indexOf(current);
+    currentOptionPosition = currentOptionPosition > -1 ? currentOptionPosition : 0;
+    const nextElementPosition = (currentOptionPosition + options.length + direction) % options.length;
+
+    options[nextElementPosition].focus();
+  }
+
+  const onKeyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    e.stopPropagation();
+    switch (e.key) {
```
AliciaSymphony, line 41: We have a enum for keys, that you can reuse
https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/blob/master/src/components/common/keyUtils.tsx

AliciaSymphony: also a `cancelEvent` util method :)

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+    const options = (Array.from(current.parentElement.getElementsByClassName('tk-dropdown-menu__item'))) as HTMLDivElement[];
+    let currentOptionPosition = options.indexOf(current);
+    currentOptionPosition = currentOptionPosition > -1 ? currentOptionPosition : 0;
+    const nextElementPosition = (currentOptionPosition + options.length + direction) % options.length;
+
+    options[nextElementPosition].focus();
+  }
+
+  const onKeyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    e.stopPropagation();
```
AliciaSymphony, line 40: Do we always want to stop the propagation? Not only when ArrowDow/Up/Enter?

liuting1014: I think so. You don't want to trigger other events on the parent, right?

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: `cancelEvent` sounds a bit confusing, sounds like you are gonna cancel the fired event  😅.
Personally I think it's clearer to just have `e.stopPropagation`, plus it's not really necessary here to have `e.preventDefault()`here.

src/components/dropdown-menu/DropdownMenu.tsx
AliciaSymphony: For (dropdown) list, W3.org suggests
`<ul><li></li>...</ul>`
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

liuting1014: The thing is we are not really creating a dropdown here even though we call it dropdown menu.  I remember there used to be a real dropdown in UIToolkit-components. But the styles did not align with the other dropdowns in Symphony. That's why we had to create a new dropdown menu component. And I think the styles were not intended to work on list or buttons...

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+      focusNextOption(e.currentTarget, -1);
+      break;
+    case Keys.ENTER:
+      onActivate(e);
+    }
+  }
+
   return (
-    <div {...rest} className={classes} onClick={onClick}>
+    <div {...rest} className={classes} onClick={onClick || onActivate} ref={forwardRef} onKeyDown={onKeyDownHandler} tabIndex={-1}>
```
AliciaSymphony, line 54: I think that we should keep binding onClick={onClick}

src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -14,18 +14,44 @@ interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
...
+    e.stopPropagation();
+    switch (e.key) {
+    case Keys.ARROW_DOWN :
+      focusNextOption(e.currentTarget, 1);
+      break;
+    case Keys.ARROW_UP:
+      focusNextOption(e.currentTarget, -1);
+      break;
+    case Keys.ENTER:
+      onActivate(e);
```
AliciaSymphony, line 49: For my understanding: Do we expect a different behaviour when we press Enter on the item, and when we click?

src/components/dropdown-menu/DropdownMenu.tsx
AliciaSymphony: If we stop propagate to the parent what we don't handle, the native behaviour of such Key "Tab", "Escape"... won't behave anymore, I wanted to double check if it is expected

src/components/dropdown-menu/DropdownMenu.tsx
AliciaSymphony: ok!

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: No, they are supposed to do the same thing.

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: Good point. I tried it just now and it does work, which is a bit weird (or maybe just something I don't have knowledge about..).

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: Since both clicking and pressing enter do the same thing, ideally we only want one handler "onActivate", doesn't matter whether the user clicked or pressed enter. So ideally we want `onClick={onActivate}` as well. I just kept the onClick for backward compatibility. In the future I think you can pass just `onActivate` and it will be called no matter you click or press enter.

symphony-annaLlorens:
Could you please add a video showing your changes on the PR?

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

symphony-annaLlorens (approved):

src/components/dropdown-menu/DropdownMenu.tsx
AliciaSymphony: ok!
Then if there are no specification for the others keys, I do think that we should not stop propagate them, and only at the case we do something (Enter, Down, Up)

src/components/dropdown-menu/DropdownMenu.tsx
liuting1014: The reason why I did not change the type of onClick was that onClick sounds like a mouse event so it might be confusing to add the KeyboardEvent type to it. But I can change it if that makes the code simpler.

spec/components/dropdown-menu/DropdownMenu.spec.tsx
```diff
@@ -50,5 +51,36 @@ describe('DropdownMenu', () => {
...
+
+    userEvent.type(screen.getByTestId('option_2'), specialChars.arrowUp);
+    expect(screen.getByTestId('option_1')).toHaveFocus();
+
+    // reaches the end
+    userEvent.type(screen.getByTestId('option_1'), specialChars.arrowUp);
+    expect(screen.getByTestId('option_2')).toHaveFocus();
+  })
+
+  it('should call onActivate when pressing enter on an option', () => {
```
AliciaSymphony, line 75: onClick

AliciaSymphony (approved):
LGTM, great work!

PR #295, opened 2021-09-20
Timeline component

nolan-symphony:
Screen Shot 2021-09-20 at 09 04 15

Screen Shot 2021-09-20 at 09 04 33

Screen Shot 2021-09-20 at 09 04 48


axeleriksson147:
"As this component is not yet integrated to the SDS I think we should wait before merge it." - Are they designing this component in SDS right now and it'll be merged in a bit?

axeleriksson147:
Could we add some tests? Timeline, expandAll, closeAll, isExpanded.

nolan-symphony:
> "As this component is not yet integrated to the SDS I think we should wait before merge it." - Are they designing this component in SDS right now and once integrated the idea is to merge this?

Yes, that's what I understood. Need to confirm with @symphony-adnane

spec/components/timeline/Timeline.spec.tsx
```diff
@@ -0,0 +1,165 @@
+import * as React from 'react';
```
axeleriksson147, line 1: Look at the documentation in `contributing.md`:

`- Even if Enzyme exists we'll use from now on only Testing Library`

symphony-jean-michael (approved):
Thank you @nolan-symphony for your contribution! 😃 
LGTM 👍 , just one comment (same as @axeleriksson147) if you can use the Testing Library instead of Enzyme that would be great. Thank you

spec/components/timeline/Timeline.spec.tsx
nolan-symphony: Done

spec/components/timeline/Timeline.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+      itemHeaderRenderer: Header,
+    }));
+    const { Timeline, timelineProps, isAllExpanded } = result.current;
+
+    expect(isAllExpanded).toEqual(false);
+    const {container} = render(<Timeline {...timelineProps}/>);
+    expect(container.firstChild).toHaveClass('tk-timeline');
+  });
+
+  it('isAllExpanded', () => {
```
axeleriksson147, line 64: We should try and test what the user sees instead of some property. Can we make sure all the items are visible when isAllExpanded is set?

spec/components/timeline/Timeline.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const { Timeline, timelineProps } = result.current;
+    let component = render(<Timeline {...timelineProps}/>);
+    expect(component.container.firstChild).toHaveClass('tk-timeline');
+    expect(result.current.isAllExpanded).toEqual(false);
+    let body = await component.queryAllByTestId('body');
+    expect(body.length).toBe(0);
+
+    expect(result.current.isAllExpanded).toEqual(false);
+
+    act(() => {
```
axeleriksson147, line 97: Can we test if expandAll works by clicking the button?

src/components/timeline/interfaces.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface UseTimeLineResult<T> {
+  isAllExpanded: boolean;
+  collapseAll: () => void;
+  expandAll: () => void;
+  timelineProps: TimelineProps<T>;
+  Timeline: React.FC<TimelineProps<T>>;
+}
+
+export interface TimelineProps<T> {
```
axeleriksson147, line 24: Should this one be able to accept React.HTMLProps so all those props can be passed to the component?

nolan-symphony: Yes why not if we want to customize the component.

spec/components/timeline/Timeline.spec.tsx
nolan-symphony: I added a test 👍

spec/components/timeline/Timeline.spec.tsx
nolan-symphony: I updated the test and now I verify that all items are visible

PR #302, opened 2021-09-29
APP-4546 Bug fixes

symphony-annaLlorens:
### Fixes:

- Loading button fixed size
- Right icon well displayed
- Added button role

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -41,20 +41,9 @@ export const Button: React.FC<ButtonProps> = ({
     prefix,
     `${prefix}--${variant}`,
     iconButton && `${prefix}--icon`,
-    iconRight && `${prefix}--icon-right`,
-    iconLeft && `${prefix}--icon-left`,
-    `${loading ? 'loading' : ''}`,
+    loading && 'loading',
     `${prefix}--${size}`
   );
-
-  const [width, setWidth] = React.useState(0);
-  const ref = React.useRef(null);
-  React.useEffect(() => {
-    if (ref.current && ref.current.getBoundingClientRect().width) {
-      setWidth(ref.current.getBoundingClientRect().width);
-    }
- 
-  }, [children]);
 
   if(variant === 'destructive') {
     console.warn('The button variant: \'destructive\' will be deprecated.\n Please use: \'primary-destructive\' instead')
@@ -62,16 +51,18 @@ export const Button: React.FC<ButtonProps> = ({
 
   return (
     <button
-      ref={ref}
       role="button"
       className={classes}
       disabled={loading || disabled}
       /* eslint-disable react/button-has-type */
       type={type}
-      style={loading && children ? {width: `${width}px`}: {}}
       {...rest}
     >
-      {loading ? <i className="animate-spin tk-icon-loading" /> : iconLeft ? <>{iconLeft}{children}</> : iconRight ? <>{children}{iconRight}</> : children}
+      {loading && <i className="animate-spin tk-icon-loading" />}
+      <span className={classNames(iconRight && `${prefix}--icon-right`,iconLeft && `${prefix}--icon-left`)}
+        style={{ visibility: loading ? 'hidden' : null }}>
+        {iconLeft}{children}{iconRight}
+      </span>
     </button>
   );
 };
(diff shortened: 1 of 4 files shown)


src/components/button/Button.tsx
```diff
@@ -46,18 +46,32 @@ export const Button: React.FC<ButtonProps> = ({
     `${loading ? 'loading' : ''}`,
     `${prefix}--${size}`
   );
-  if(variant==='destructive') {
+
+  const [width, setWidth] = React.useState(0);
+  const ref = React.useRef(null);
+  React.useEffect(() => {
+    if (ref.current && ref.current.getBoundingClientRect().width) {
```
symphony-adnane, line 53: can't we do this with CSS only? 
What are we trying to achieve?

symphony-annaLlorens: To get dinamically the width of the button, this way when the button is loading it keep the same size

symphony-annaLlorens: To get dynamically the width of the button, this way when the button is loading it keep the same size

AliciaSymphony: I think that you can put the text to hidden, and center the loading icon with absolute

src/components/button/Button.tsx
```diff
@@ -46,18 +46,32 @@ export const Button: React.FC<ButtonProps> = ({
...
+      role="button"
       className={classes}
       disabled={loading || disabled}
       /* eslint-disable react/button-has-type */
       type={type}
+      style={loading && children ? {width: `${width}px`}: {}}
       {...rest}
     >
-      {loading ? <i className="animate-spin tk-icon-loading" /> : iconLeft ? <>{iconLeft}{children}</> : iconRight ? <>{iconRight}{children}</> : children}
+      {loading ? <i className="animate-spin tk-icon-loading" /> : iconLeft ? <>{iconLeft}{children}</> : iconRight ? <>{children}{iconRight}</> : children}
```
AliciaSymphony, line 74: You can simply put

```
{loading ? <i className="animate-spin tk-icon-loading" /> :
  <i className="animate-spin tk-icon-loading" /> :  <>{iconLeft ? iconLeft : null}{children}{iconRight ? iconRight : null}</>
}
```

I don't think that we should restrict to left or right, if the user want to put both?

src/components/button/Button.tsx
symphony-annaLlorens: done

src/components/button/Button.tsx
symphony-annaLlorens: done

src/components/button/Button.tsx
```diff
@@ -41,23 +41,28 @@ export const Button: React.FC<ButtonProps> = ({
...
   );
-  if(variant==='destructive') {
+
+  if(variant === 'destructive') {
     console.warn('The button variant: \'destructive\' will be deprecated.\n Please use: \'primary-destructive\' instead')
   }
+
   return (
     <button
+      role="button"
```
symphony-jean-michael, line 54: Just curious. Why do you need to add the role 'button' on a button tag? (See https://stackoverflow.com/questions/26934719/do-i-need-role-button-on-a-button)

src/components/button/Button.tsx
```diff
@@ -41,23 +41,28 @@ export const Button: React.FC<ButtonProps> = ({
...
       className={classes}
       disabled={loading || disabled}
       /* eslint-disable react/button-has-type */
       type={type}
       {...rest}
     >
-      {loading ? <i className="animate-spin tk-icon-loading" /> : iconLeft ? <>{iconLeft}{children}</> : iconRight ? <>{iconRight}{children}</> : children}
+      {loading && <i className="animate-spin tk-icon-loading" />}
+      <span className={classNames(iconRight && `${prefix}--icon-right`,iconLeft && `${prefix}--icon-left`)}
```
symphony-jean-michael, line 62: IMO, classNames should take an object in parameter, isn't it ?
```classNames({ iconRight: `${prefix}--icon-right`, iconLeft: `${prefix}--icon-left` })```
(Se: https://www.npmjs.com/package/classnames)

AliciaSymphony (approved):

symphony-jean-michael (approved):

PR #303, opened 2021-09-29
APP-4531 Add missing descriptions in Docs tab of StoryBook

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4531

**Changes**
Add missing dependency to 'react-docgen-typescript-loader' package because the descriptions were not correctly loaded from our components.
- Update TimePicker description:
- Code cleanup in other stories

## TimePicker

**Demo without the fix**
Screenshot 2021-09-29 at 16 06 57

**Demo with the fix**
Screenshot 2021-09-29 at 16 05 30

## Other components, for example: Button

**Demo without the fix**
Screenshot 2021-09-29 at 16 11 15

**Demo with the fix**
Screenshot 2021-09-29 at 16 11 29

### DIFF
--- a/src/components/time-picker/TimePicker.tsx
+++ b/src/components/time-picker/TimePicker.tsx
@@ -1,12 +1,10 @@
 import { Dropdown, DropdownOption } from '../dropdown';
-import { HasValidationProps } from '../validation/interfaces';
-import { HasTooltipProps } from '../tooltip/interfaces';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import { useEffect, useMemo, useState } from 'react';
 import { Keys } from '../common/keyUtils';
 import { ErrorMessages } from '../validation/interfaces';
-import { DisabledTime, TimePickerValue } from './interfaces';
+import { DisabledTime, TimePickerProps } from './interfaces';
 import {
   formatTimeISO,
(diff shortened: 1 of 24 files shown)


AliciaSymphony (approved):
Nice!

src/components/time-picker/TimePicker.tsx
```diff
@@ -434,4 +415,13 @@ const handleFocus = (event) => {
   }
 };
 
+TimePicker.defaultProps = {
+  disabledTimes: [],
+  format: getUserFormat(),
+  min: '00:00:00',
+  max: '23:59:59',
+  step: STEP.DEFAULT_STEP_VALUE,
```
symphony-adnane, line 423: is step an enum?

symphony-adnane (approved):

src/components/time-picker/TimePicker.tsx
symphony-jean-michael: Yes it is

symphony-jean-michael: The default value is taken from an enum, but it is a number in seconds. (min: 600, default: 900, max 43200). The step value can be anything between min and max

PR #312, opened 2021-10-18
APP-4602 Removes unnecessary span tag when no icon

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4602

**Changes**
- Lint file (I created an improvement ticket to add a Husky pre-commit hook to auto lint the files: https://perzoinc.atlassian.net/browse/APP-4613 )
- Don't generate the unnecessary span tag when the is no iconRigh/iconLeft

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -7,7 +7,7 @@ const prefix = 'tk-button';
 export type ButtonProps = {
   /** If true, add an Icon component as children */
   iconButton?: boolean;
-   /** Content of the button*/
+  /** Content of the button*/
   children?: React.ReactNode;
   /** Optional CSS class name */
   className?: string;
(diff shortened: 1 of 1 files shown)


src/components/button/Button.tsx
```diff
@@ -58,10 +58,10 @@ export const Button: React.FC<ButtonProps> = ({
       {...rest}
     >
       {loading && <i className="animate-spin tk-icon-loading" />}
-      <span className={classNames({ [`${prefix}--icon-right`] : iconRight , [`${prefix}--icon-left`]: iconLeft })}
+      {iconLeft || iconRight ? (<span className={classNames({ [`${prefix}--icon-left`]: iconLeft, [`${prefix}--icon-right`]: iconRight })}
         style={{ visibility: loading ? 'hidden' : null }}>
         {iconLeft}{children}{iconRight}
-      </span>
+      </span>) : children}
```
AliciaSymphony, line 64: be careful we loose the logic of the loading visibility in the else

PR #324, opened 2021-10-28
APP-4551 - feat(date-picker): Date Picker support rendering inside po…

AliciaSymphony:
…rtal
## Ticket
https://perzoinc.atlassian.net/browse/APP-4551

## Description
- Give the possibility to the Date Picker to render in a portal
- Give the possibility to the Date Picker to have additional css style for this portal
- Give the possibility to the block the scroll when the Date Picker is open
- ⚠️ `showOverlay` and `menuShouldBlockScroll` are incompatible

Keep the prefix "menu" to be aligned with the TK Dropdown `menuPortalStyles?: CSSProperties;  menuPortalTarget?: HTMLElement;  menuShouldBlockScroll?: boolean;`

## Demo
Oct-28-2021 18-21-21

Nov-03-2021 12-40-08


src/components/common/keyUtils.tsx
```diff
@@ -19,3 +19,15 @@ export function cancelEvent(e) {
   e.preventDefault();
   e.stopPropagation();
 }
+
+export function getScrollParent(node) {
+  if (node == null) {
```
symphony-jean-michael, line 24: In cases where the node is undefined, I suggest using the following syntax to avoid using a non-strict comparison
```
if(!node){
   return null;
}
```

src/components/date-picker/DatePicker.tsx
```diff
@@ -276,12 +287,33 @@ class DatePicker extends Component<
...
       ],
     });
+
+
+    if (menuShouldBlockScroll) {
+      const scrollContainer = getScrollParent(refContainer);
+      if (scrollContainer) {
+        scrollContainer.addEventListener('touchmove',this.handleScrollParent);
+        scrollContainer.addEventListener('DOMMouseScroll',this.handleScrollParent);
+        scrollContainer.addEventListener('mousewheel',this.handleScrollParent);
```
symphony-jean-michael, line 297: Declare 'touchmove', 'DOMMouseScroll', mousewheel' in some const/enum in a utility file (Like keyUtils.tsx)

src/components/date-picker/DatePicker.tsx
```diff
@@ -242,19 +253,19 @@ class DatePicker extends Component<
     }
   }
 
-  private setRefContainer(node) {
+  private setRefContainer(node: HTMLDivElement) {
```
symphony-jean-michael, line 256: 👍

src/components/date-picker/DatePicker.tsx
```diff
@@ -276,12 +287,33 @@ class DatePicker extends Component<
...
+    const { menuShouldBlockScroll } = this.props;
+    const { refContainer } = this.state;
+
     if (this.dayPickerInstance) {
       this.dayPickerInstance.destroy();
       Reflect.deleteProperty(this, 'dayPickerInstance');
+      if (menuShouldBlockScroll) {
+        const scrollContainer = getScrollParent(refContainer);
+        if (scrollContainer) {
+          scrollContainer.removeEventListener('touchmove',this.handleScrollParent);
```
symphony-jean-michael, line 312: FYI, React-Select uses the lib https://www.npmjs.com/package/react-scrolllock
See commit: https://github.com/JedWatson/react-select/commit/d2eea09e5bd046a9481513a68c84d733c5ea9644
This library also handles mobile devices. This could be a good solution to use it. What do you think about it?

stories/Dropdown.stories.tsx
```diff
@@ -248,7 +248,7 @@ export const Select: React.FC = () => (
 
 export const Portal = PortalTemplate.bind({});
 Portal.args = {
-  title: 'A Dropdown Menu in a Portal',
+  title: 'A Dropdown rendering its menu inside a Portal',
```
symphony-jean-michael, line 251: 👍

src/components/date-picker/DatePicker.tsx
AliciaSymphony: Thank you for sharing,
react-scrolllock is made to block the scroll on <body>, but I am blocking the scroll of a specific div

src/components/date-picker/DatePicker.tsx
AliciaSymphony: ok

symphony-jean-michael (approved):

symphony-adnane (approved):
Looks good to me ! 
If I understood well though, we block the scroll on the direct parent, how would this behave is the one that's scrollable is the parent of the parent? 
ex: a datepicker inside an old card, inside a chat (the scrollable) -> we still want the chat to not scroll

PR #329, opened 2021-11-09
APP-4672 Improve input validator

symphony-annaLlorens:
### Description
- Fix required validator to Dropdown
- Increase test cases to Input validators

### Before
Screenshot 2021-11-09 at 14 14 33

### After fix

Screenshot 2021-11-09 at 14 15 16

[Jira-ticket](https://perzoinc.atlassian.net/browse/APP-4672)

### DIFF
--- a/spec/core/validators/validators.spec.tsx
+++ b/spec/core/validators/validators.spec.tsx
@@ -51,7 +51,7 @@ describe('Input Validators', () => {
     expect(await Validators.EmptyString(emptyArray)).toEqual({ emptyString: true });
   });
 
- it('should validate "minlength" Input correctly', async () => {
+  it('should validate "minlength" Input correctly', async () => {
     expect(await Validators.MinLength(3)('aaa')).toEqual(null);
     expect(await Validators.MinLength(3)(null)).toEqual({ minlength: true });
     expect(await Validators.MinLength(3)(undefined)).toEqual({ minlength: true });
@@ -61,7 +61,7 @@ describe('Input Validators', () => {
     expect(await Validators.MinLength(3)(emptyArray)).toEqual( {minlength: true});
     expect(await Validators.MinLength(3)(arrayWithValues)).toEqual( {minlength: true});
     expect(await Validators.MinLength(1)(arrayWithValues)).toEqual(null);
-
+  });
   it('should validate "maxLength" Input correctly', async () => {
     expect(await Validators.MaxLength(3)('abcd')).toEqual({ maxLength: true });
     expect(await Validators.MaxLength(3)(stringWithValue)).toEqual(null);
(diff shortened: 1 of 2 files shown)


src/core/validators/validators.ts
```diff
@@ -119,8 +121,8 @@ const Email: ValidatorFn = (value) => {
   return Promise.resolve(null);
 };
 
-const isEmpty = (value: string | any[]) => {
-  return !value?.length || (typeof value==='string' && value?.trim?.() === '');
+const isEmptyValue = (value: Value) => {
+  return !value || (typeof value==='string' && value?.trim?.() === '') || Object.keys(value).length === 0;
```
symphony-adnane, line 125: This won't work for an empty array

symphony-adnane: could combine two first as : !value?.trim?()

src/core/validators/validators.ts
```diff
@@ -57,7 +59,7 @@ const MinLength = (minlength: number): ValidatorFn => {
  */
 const MaxLength = (maxLength: number): ValidatorFn => {
   return (value) => {
-    if (!value|| maxLength >= value.length) {
+    if (!value|| maxLength >= value.length || Object.getPrototypeOf(value) === Object.prototype) {
```
symphony-adnane, line 62: you could use lodash isEmpty for all these checks imo

src/core/validators/validators.ts
```diff
@@ -31,7 +33,7 @@ const EmptyString: ValidatorFn = (value) => {
   console.warn(
     'Calling a deprecated validator (EmptyString), please use the Required validator instead'
   );
-  if (isEmpty(value)) {
+  if (isEmptyValue(value)) {
```
symphony-adnane, line 36: https://lodash.com/docs/4.17.15#isEmpty

src/core/validators/validators.ts
symphony-annaLlorens: Here, what I'm checking is that is an object strict. If I use the lodash ._isObject() it does not differentiate [] from {}

src/core/validators/validators.ts
symphony-annaLlorens: isEmptyValue() is also checking that the string doesn't contain spaces. 
The isEmpty() from lodash is not checking it.

symphony-adnane: you may wanna trim it if possible before passing it ...

spec/core/validators/validators.spec.tsx
```diff
@@ -1,49 +1,82 @@
...
     expect(await Validators.MaxLength(3)('abcd')).toEqual({ maxLength: true });
+    expect(await Validators.MaxLength(3)(stringWithValue)).toEqual(null);
+    expect(await Validators.MaxLength(3)(null)).toEqual(null);
+    expect(await Validators.MaxLength(3)(undefined)).toEqual(null);
+    expect(await Validators.MaxLength(3)(emptyString)).toEqual(null);
+    expect(await Validators.MaxLength(3)(objectWithValue)).toEqual(null);
+    expect(await Validators.MaxLength(3)(emptyObject)).toEqual(null);
+    expect(await Validators.MaxLength(3)(emptyArray)).toEqual(null);
+    expect(await Validators.MaxLength(2)(arrayWithValues)).toEqual(null);
+    expect(await Validators.MaxLength(1)(arrayWithValues)).toEqual( {maxLength: true});
```
symphony-adnane, line 75: are minlength 0 and maxlength 0 something that should exist? or should we throw an exception?

symphony-adnane (approved):

spec/core/validators/validators.spec.tsx
symphony-annaLlorens: Added an exception

AliciaSymphony (approved):

PR #331, opened 2021-11-11
RTC-11877 - Not render a tooltip in TextEllipsis if tooltipProps.description or children are undefined

axeleriksson147:
If the child is undefined, the TextEllipsis returns an error because description cannot be empty for tooltip.

### DIFF
--- a/src/components/text-ellipsis/TextEllipsis.tsx
+++ b/src/components/text-ellipsis/TextEllipsis.tsx
@@ -48,12 +48,16 @@ export const TextEllipsis: React.FC<TextEllipsisProps> = ({
       { children }
     </span>)
   }
+
+  if(!tooltipProps?.description && !children) {
+    return null
+  }
   
   if(tooltipOnEllipsis) {
     return(
       <Tooltip
         {...tooltipProps}
-        description={ tooltipProps?.description ?? children as JSX.Element }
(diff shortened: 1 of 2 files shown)


src/components/text-ellipsis/TextEllipsis.tsx
```diff
@@ -48,12 +48,16 @@ export const TextEllipsis: React.FC<TextEllipsisProps> = ({
       { children }
     </span>)
   }
+
+  if(!tooltipProps?.description && !children) {
```
symphony-annaLlorens, line 52: The description is a required prop on the Tooltip component, why we need to check it here?

axeleriksson147: tooltipProps is `Partial<Partial<TooltipProps>` so any value could be undefined.

`TextEllipsis` works so that it will by default pass the `props.children` as the description for `Tooltip`, unless you override it with `tooltipProps.description`.

Instead of having to add checks for if your `children` or `tooltipProps.description` are undefined I think it's better if we can pass that to TextEllipsis and it handles it internally.

symphony-annaLlorens (approved):
LGTM 👍

symphony-jean-michael (approved):
LGTM. Thank you for your contribution :)

PR #337, opened 2021-11-22
fix: unwanted behaviour when the menu is not displayed

AnthonyDS-symphony:
For the need of [this CES ticket](https://perzoinc.atlassian.net/browse/CES-5147), I need to replace our DropDownMenu with the one from the UI Toolkit.

When implementing it, I found an issue because event listener `click` triggered the `mouseEventHandler` event for each click every where on the page **even if the menu is not open**.

So, when you click on a button that should open the menu, the `close()` is automatically triggered in the same time: so we can't open the menu.

I've just added a watch on `show` property in order to skip event listeners.
In short:
- `show=false`: we don't need to listen events
- `show=true`: we need to listen events

### DIFF
--- a/spec/components/dropdown-menu/DropdownMenu.spec.tsx
+++ b/spec/components/dropdown-menu/DropdownMenu.spec.tsx
@@ -36,6 +36,20 @@ describe('DropdownMenu', () => {
     expect(close).toHaveBeenCalled();
   });
 
+  it('shouldn\'t call onClose when clicking outside if show=false', () => {
+    const close = jest.fn();
+    render(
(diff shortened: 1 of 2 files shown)


liuting1014 (approved):

liuting1014:
It would be nice to add some unit tests :)

AliciaSymphony (approved):
Great improvement, thank you for your contribution!!

symphony-jean-michael (approved):
LGTM. Thank you for the fix @AnthonyDS-symphony ! 😃

PR #342, opened 2021-11-30
fix: Show Code button is now accessible

AnthonyDS-symphony:
Improving Modal component docs (all stories).
I've just added `StoryBackground` component to reduce the `height` only for Storybook because the modal's backdrop was over the `Show Code` button from the Docs.

Now we can click on it and see how component works.

Sample before:
image

Sample after:
image
image


AliciaSymphony (approved):
Really appreciated, 💪 you rock!!

symphony-jean-michael (approved):
Thank you! @AnthonyDS-symphony :)

PR #359, opened 2021-12-10
Add auto pre-commit check

Symphony-cyril:


.eslintcache
```diff
@@ -0,0 +1 @@
+[{"/Users/cyril.ruffa/Documents/Symphony/symphony-bdk-ui-toolkit-components/src/index.ts":"1"},{"size":48,"mtime":1639072644512}]
```
symphony-adnane, line 1: could you delete this file and push again? ;)

AliciaSymphony, line 1: Looks like it was pushed even though it is in your .gitignore

Symphony-cyril: yes i ignore why i have removed it manually

PR #363, opened 2021-12-15
APP-4751 Added Icon types

symphony-annaLlorens:
### Description 
- APP-4647 Added Icon types

[Jira-ticket](https://perzoinc.atlassian.net/browse/APP-4647)

### Demo 
Screenshot 2021-12-15 at 12 39 54

### DIFF
--- a/src/components/icon/Icon.tsx
+++ b/src/components/icon/Icon.tsx
@@ -53,7 +53,7 @@ const Icon: React.FC<IconProps> = ({
 };
 
 Icon.propTypes = {
-  iconName: PropTypes.any.isRequired,
+  iconName: PropTypes.string.isRequired as PropTypes.Validator<TkIcon>,
   className: PropTypes.string,
   disabled: PropTypes.bool,
   forwardRef: PropTypes.any,
--- a/src/components/icon/Icon.tsx
+++ b/src/components/icon/Icon.tsx
@@ -1,10 +1,11 @@
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import classnames from 'classnames';
+import { TkIcon } from  '@symphony-ui/uitoolkit-styles/dist/fonts/tk-icons';
 
 export interface IconProps extends React.HTMLProps<HTMLElement> {
   /** Required; Designates a specific icon */
-  iconName: string;
+  iconName: TkIcon;
   /** Optional CSS class name */
   className?: string;
   /** If true, the icon will be disabled */
@@ -52,7 +53,7 @@ const Icon: React.FC<IconProps> = ({
 };
 
 Icon.propTypes = {
-  iconName: PropTypes.string.isRequired,
+  iconName: PropTypes.any.isRequired,
   className: PropTypes.string,
   disabled: PropTypes.bool,
   forwardRef: PropTypes.any,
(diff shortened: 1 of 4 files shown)


src/components/icon/Icon.tsx
```diff
@@ -1,10 +1,11 @@
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import classnames from 'classnames';
+import { TkIcon } from  '@symphony-ui/uitoolkit-styles/dist/fonts/tk-icons';
```
symphony-adnane, line 4: this import looks a bit weird to me :/ 
we should be able to do something like 
import {TkIcon} from '@symphony-ui/uitoolkit-styles'

stories/Icons.stories.tsx
```diff
@@ -10,7 +11,7 @@ export const Icons: React.FC = () => (
     {sortName.map((value, index) => {
       return (
         <div key={index} className="preview">
-          <Icon iconName={value}></Icon>
+          <Icon iconName={value as TkIcon}></Icon>
```
AliciaSymphony, line 14: Let's avoid "as Type" as much as we can

src/components/icon/Icon.tsx
```diff
@@ -52,7 +53,7 @@ const Icon: React.FC<IconProps> = ({
 };
 
 Icon.propTypes = {
-  iconName: PropTypes.string.isRequired,
+  iconName: PropTypes.any.isRequired,
```
symphony-jean-michael, line 56: We should try to avoid using 'any'. I didn't find any good solution for using 'type' with PropTypes.
The only solution is to use:
```
  iconName: PropTypes.string.isRequired as PropTypes.Validator<TkIcon>,
```
With this line, we use at least the 'string' Proptype (and there is no error thanks to the `as PropTypes.Validator<TkIcon>`)

src/components/icon/Icon.tsx
symphony-jean-michael: I agree with @symphony-adnane 
A question, maybe we should not provide them in the 'dist' folder, but at the root folder to be able to use the line suggested by @symphony-adnane 🤔

src/components/toast/Toast.tsx
```diff
@@ -81,7 +82,7 @@ Toast.propTypes = {
   className: PropTypes.string,
   closeIcon: PropTypes.bool,
   content: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
-  leftIcon: PropTypes.string,
+  leftIcon: PropTypes.any,
```
symphony-jean-michael, line 85: Same comment as above.
I suggest to use ` PropTypes.string as PropTypes.Validator<TkIcon>,`

stories/Icons.stories.tsx
symphony-annaLlorens: The name of the Icons are coming from `tk-icons.codepoints.json` and they are parsed as a string[]. T. See: https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/blob/b79b161433f64527e511b7e001af84a1f742c5b0/stories/Icons.stories.tsx#L6
They will be the same as the types on the tk-icons.ts file but typescript can not know when is parsing the file.

src/components/toast/Toast.tsx
symphony-annaLlorens: Oh very nice!, thaanks :)

AliciaSymphony (approved):

PR #365, opened 2021-12-16
APP-4735 Fix: datePicker required error displayed

symphony-annaLlorens:
### Bug
Screenshot 2021-12-16 at 15 53 14

### Fix
Screenshot 2021-12-16 at 15 56 45

### DIFF
--- a/spec/core/validators/validators.spec.tsx
+++ b/spec/core/validators/validators.spec.tsx
@@ -18,6 +18,7 @@ describe('Input Validators', () => {
     expect(await Validators.Required(stringWithValue)).toEqual(null);
     expect(await Validators.Required(objectWithValue)).toEqual(null);
     expect(await Validators.Required(arrayWithValues)).toEqual(null);
+    expect(await Validators.Required(new Date())).toEqual(null);
     expect(await Validators.Required(emptyObject)).toEqual({ required: true }); 
     expect(await Validators.Required(emptyArray)).toEqual({ required: true });
 
--- a/spec/core/validators/validators.spec.tsx
+++ b/spec/core/validators/validators.spec.tsx
@@ -19,6 +19,7 @@ describe('Input Validators', () => {
     expect(await Validators.Required(objectWithValue)).toEqual(null);
     expect(await Validators.Required(arrayWithValues)).toEqual(null);
     expect(await Validators.Required(new Date())).toEqual(null);
+    expect(await Validators.Required(3)).toEqual(null);
     expect(await Validators.Required(emptyObject)).toEqual({ required: true }); 
     expect(await Validators.Required(emptyArray)).toEqual({ required: true });
 
--- a/src/core/validators/validators.ts
+++ b/src/core/validators/validators.ts
@@ -50,11 +50,7 @@ const MinLength = (minlength: number): ValidatorFn => {
     );
   }
   return (value) => {
-    if (
-      value &&
-      (minlength <= (value as any).length ||
-        Object.getPrototypeOf(value) === Object.prototype)
-    ) {
+    if (value && (minlength <= value.length || Object.getPrototypeOf(value) === Object.prototype)) {
       return Promise.resolve(null);
     }
     return Promise.resolve({ minlength: true });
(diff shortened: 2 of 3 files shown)


src/core/validators/validators.ts
```diff
@@ -145,8 +157,10 @@ const Url: ValidatorFn = (value) => {
 };
 
 const isEmptyValue = (value: Value) => {
-  return  _.isEmpty(value) || (typeof value==='string' && value?.trim?.() === '');
-}
+  return value instanceof Date
```
symphony-adnane, line 160: Are we going to be checking each type? each time we have an issue with a new type? :D

AliciaSymphony: I agree that a generic solution should be provided
Also reading the code, `isEmptyValue` will always be 'false' is the value is a Date (?)

symphony-annaLlorens: yes, because if there is a Date defined the value is not empty. 
The `._isEmpty()` return true if you give them a date. Do you have any suggestions on how to check it without breaking other cases?

src/core/validators/validators.ts
```diff
@@ -144,9 +158,9 @@ const Url: ValidatorFn = (value) => {
   return Promise.resolve(null);
 };
 
-const isEmptyValue = (value: Value) => {
-  return  _.isEmpty(value) || (typeof value==='string' && value?.trim?.() === '');
-}
+const isEmptyValue = (value: Value) =>
+  (_.isEmpty(value) && !_.isDate(value)) ||
+  (_.isString(value) && value?.trim?.() === '');
```
symphony-adnane, line 163: if trim exists then it's a string... so second statement includes first right?

symphony-annaLlorens: yes

src/core/validators/validators.ts
```diff
@@ -144,9 +158,9 @@ const Url: ValidatorFn = (value) => {
   return Promise.resolve(null);
 };
 
-const isEmptyValue = (value: Value) => {
-  return  _.isEmpty(value) || (typeof value==='string' && value?.trim?.() === '');
-}
+const isEmptyValue = (value: Value) =>
+  (_.isEmpty(value) && !_.isDate(value)) ||
+  (_.isString(value) && value.trim() === '');
```
AliciaSymphony, line 163: Would be nice to have a generic solution to handle others types, I am thinking about numbers for instance

AliciaSymphony: _.isEmpty(value) || _.isNill(value)) ||
  (_.isString(value) && value.trim() === '')
Something like this could work?

AliciaSymphony: Ok, after some advanced research, I propose the following


```
return ((_.isArray(value) || _.isPlainObject(value)) && _.isEmpty(value)) || _.isNil(value) || (_.isString(value) && value.trim() === '');
```

spec/core/validators/validators.spec.tsx
```diff
@@ -18,6 +18,7 @@ describe('Input Validators', () => {
     expect(await Validators.Required(stringWithValue)).toEqual(null);
     expect(await Validators.Required(objectWithValue)).toEqual(null);
     expect(await Validators.Required(arrayWithValues)).toEqual(null);
+    expect(await Validators.Required(new Date())).toEqual(null);
```
AliciaSymphony, line 21: You could test also this
`expect(await Validators.Required(123546)).toEqual(null);`

src/core/validators/validators.ts
```diff
@@ -1,6 +1,10 @@
 import * as _ from 'lodash';
 
-type Value = string | { [key: string]: string } | { [key: string]: string }[];
+type Value =
+  | string
+  | { [key: string]: string }
+  | { [key: string]: string }[]
+  | Date;
```
AliciaSymphony, line 7: This type looks too specific, we cannot really predict what the user will pass as value
It can be string[], number, number[], .... or others

I am thinking the other way around, what do we **not** accept...? only Function?

symphony-annaLlorens: Yes, you are right. We should have to add each time there is a new type of value. Should it be any?

src/core/validators/validators.ts
symphony-annaLlorens: Yes, thank you :)!!  This was the idea. The `_.isEmpty` needed to be combined with other method to not give false positives.

spec/core/validators/validators.spec.tsx
symphony-annaLlorens: We could use the library from the React confonrence to generate random values to test :D

symphony-jean-michael (approved):

AliciaSymphony (approved):

PR #330, opened 2021-11-09
APP-4603  Pass onCopy/onCut/onDrag on Dropdown component

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4603

**Changes**
-TimePicker: Pass the `onCopy`/`onCut`/`onDrag` props to the Dropdown component
-Dropdown: Apply `onCopy`/`onCut`/`onDrag` props on the Input if they are defined

(FYI, I had to use 'any' in the `isOptionDisabled` and `isOptionSelected` props because I got a TypeScript error. In a major UI-TK version, we will replace the `any` with the right type.)

**Demo**
In NEXUS, it fixes the prevent copy feature

https://user-images.githubusercontent.com/66668470/143472147-695f583a-0a2e-46b0-9834-35ea9abf2a6d.mov

PR #366, opened 2021-12-20
APP-4755 Fix overlapping interfaces for the value property

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4755

**Changes**
A regression was introduced with PR #330. The 'value' prop type was overlapped by the HTMLInput interface.

**Before fix**
Screenshot 2021-12-20 at 16 36 07

**After fix:**
Screenshot 2021-12-20 at 16 35 11

### DIFF
--- a/src/components/dropdown/interfaces.ts
+++ b/src/components/dropdown/interfaces.ts
@@ -7,6 +7,11 @@ export interface OptionRendererProps<T> {
   data: T;
 }
 
+export type HTMLInputProps = Omit<
+  React.HTMLProps<HTMLInputElement>,
+  'onChange' | 'value'
+>;
+
 export interface TagRendererProps<T> extends OptionRendererProps<T> {
   remove: () => any;
 }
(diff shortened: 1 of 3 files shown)


src/components/dropdown/interfaces.ts
```diff
@@ -170,7 +170,7 @@ export type DropdownProps<T> = {
   termSearchMessage?: ((term: string) => string) | string;
   /** Color variant of the dropdown */
   variant?: 'destructive';
-} & Omit<React.HTMLProps<HTMLInputElement>, 'onChange'> &
+} & Omit<React.HTMLProps<HTMLInputElement>, 'onChange' | 'value'> &
```
AliciaSymphony, line 173: Let's refactor this logic, in a `HTMLInputProps = Omit<React.HTMLProps<HTMLInputElement>, 'onChange' | 'value'>` for instance?

symphony-annaLlorens (approved):

src/components/dropdown/interfaces.ts
```diff
@@ -7,6 +7,11 @@ export interface OptionRendererProps<T> {
   data: T;
 }
 
+export type HTMLInputProps = Omit<
+  React.HTMLProps<HTMLInputElement>,
+  'onChange'
+>;
```
AliciaSymphony, line 13: Looks like we lost the 'value' from `Omit<React.HTMLProps<HTMLInputElement>, 'onChange' | 'value'>` or did I miss something 😂

AliciaSymphony (approved):
LGTM

src/components/dropdown/interfaces.ts
symphony-adnane: the value is already part of our props @AliciaSymphony (check multimodeprops and singlemodeprops)

PR #371, opened 2021-12-21
APP-4710 Remove 'any' from the TimePicker component

symphony-jean-michael:
**Jira ticket**
https://perzoinc.atlassian.net/browse/APP-4710

**Changes**
- Remove 'any' typescript type from the TimePicker component.
   **Breaking changes:**
    -  **TimePicker**: Use TimePickerOption instead of 'any'
    -  **Dropdown**: `isOptionDisabled` and `isOptionSelected` methods will be called now with the entire option and not the subset `option.data`.
 
- Update browserslist dependency


src/components/dropdown/Dropdown.tsx
```diff
@@ -99,11 +99,11 @@ export class Dropdown<T = LabelValue> extends React.Component<
   };
 
   handleIsOptionDisabled = this.props.isOptionDisabled
-    ? (option: any) => this.props.isOptionDisabled(option.data)
+    ? (option: T) => this.props.isOptionDisabled(option)
```
AliciaSymphony, line 102: If I am not mistaken, we still need to pass option.data to `this.props.isOptionDisabled` (?)

AliciaSymphony (approved):
I am a bit concerned by the breaking change of "KeyboardEvent", "MouseEvent", "FocusEvent" type of some props

src/components/dropdown/Dropdown.tsx
symphony-jean-michael: Previously, we assumed that there was a 'data' object in the option, but it was not always the case (If for example, the user instantiate a dropdown with Dropdown<HisType>)
Now, we send the full option and not just a subset of the option. This was used in the TimePicker, I updated the code. And I'm going to list this in the Breaking changes of the v3

symphony-jean-michael:
> I am a bit concerned by the breaking change of "KeyboardEvent", "MouseEvent", "FocusEvent" type of some props

@AliciaSymphony I am taking advantage of the major version to remove 'any', this may introduce breaking changes but I will list them in the release note.

src/components/dropdown/Dropdown.tsx
AliciaSymphony: Ok we are not only replacing 'any' but also changing the type expected by the prop. This will directly impact NEXUS (and other consumers)

PR #372, opened 2021-12-22
UIToolkit improvements

symphony-annaLlorens:
### Description 
- Omit `defaultValue` and `size` from 'HTMLInputElements' props
- Update dropdown validator storybook example
- Use `TKIcon` in customRender tag storybook


src/components/dropdown/interfaces.ts
```diff
@@ -9,7 +9,7 @@ export interface OptionRendererProps<T> {
 
 export type HTMLInputProps = Omit<
   React.HTMLProps<HTMLInputElement>,
-  'onChange' | 'value'
+  'onChange' | 'value' | 'size' | 'defaultValue'
```
symphony-adnane, line 12: are these conflicting also?

symphony-jean-michael, line 12: If there are blocking issues, then I suggest you add this line also in the 2.11.2-branch. (To include it in 2.11.2 TK release)

symphony-jean-michael (approved):

src/components/dropdown/interfaces.ts
symphony-annaLlorens: yes, same issue

symphony-annaLlorens: `size` and `defaultValue` where not part of the Dropdown interface at branch 2.11.2

AliciaSymphony: I am wondering if this was also the root cause of this issue https://perzoinc.atlassian.net/browse/APP-4637 😂

AliciaSymphony: `defaultValue` is here since (almost?) the beginning of the component
but I agree that `size` was introduced by the feature branch SDS-INPUT-GROUP (=not released yet)

symphony-annaLlorens: I just checked, `defaultValue` is needed but there were no examples on the storybook :/

AliciaSymphony: There is one here https://github.com/SymphonyPlatformSolutions/symphony-bdk-ui-toolkit-components/blob/master/stories/Dropdown.stories.tsx#L226

AliciaSymphony (approved):

PR #381, opened 2021-12-27
APP-4654 feat(SelectionInput): accept node as label

AliciaSymphony:
Prevent to have the following warning when passing a node to the label prop
```
"Warning: Failed prop type: Invalid prop `label` of type `object` supplied to `Checkbox`, expected `string`.
```

### DIFF
--- a/spec/components/selection/SelectionInput.spec.tsx
+++ b/spec/components/selection/SelectionInput.spec.tsx
@@ -1,4 +1,6 @@
 import { mount, shallow } from 'enzyme';
+import { render } from '@testing-library/react';
+
 import * as React from 'react';
 import SelectionTypes from '../../../src/components/selection/SelectionTypes';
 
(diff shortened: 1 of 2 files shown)


symphony-annaLlorens (approved):

symphony-annaLlorens:
Fix lint issue

spec/components/selection/SelectionInput.spec.tsx
```diff
@@ -48,6 +50,18 @@ describe('SelectionInput Component', () => {
       expect(wrapper.find('input').prop('value')).toEqual(props.value);
     });
 
+    it('render a Radio with node label', () => {
```
symphony-annaLlorens, line 53: Could we also check label as string?

AliciaSymphony: indeed!

symphony-jean-michael (approved):
LGTM

PR #421, opened 2022-01-14
Time Picker - Fix null selectedOption

AliciaSymphony:
## Description
The selectedOption was set to null

## Solution
Fix access to hours, minutes, seconds since the recent refactor

### Observed
image (14)

### Expected
image


symphony-jean-michael (approved):
Thank you for the fix @AliciaSymphony 👍

PR #451, opened 2022-02-11
APP-4919: Fix displayed option label in Dropdown (multi select)

antoinerollindev:
**APP-4919: Fix displayed option label in Dropdown (multi select)**

https://perzoinc.atlassian.net/browse/APP-4919

When using `Dropdown` (with `isMultiSelect` being `true`), and providing options that does not include a `label` attribute, the option labels are not displayed.

- adds `bindLabel` prop in `Dropdown`, so we can define a way to get the label from an option (default being `(option) => option.label`)
- plugs this `bindLabel` prop to the `getOptionLabel` prop of the `DropdownTag`
- updates the `MultiValue` custom renderer so that it calls `getOptionLabel(option)` instead of `option.label`
- adds related tests

-----

**Before**

https://user-images.githubusercontent.com/66251236/153578986-201aa3a6-f028-443e-9d53-1f038ca8db67.mov

-----

**After**

https://user-images.githubusercontent.com/66251236/153579018-4ab3c8c5-e3cb-4c46-988c-a3e6583ba725.mov

### DIFF
--- a/packages/components/spec/components/dropdown/Dropdown.spec.tsx
+++ b/packages/components/spec/components/dropdown/Dropdown.spec.tsx
@@ -108,6 +108,39 @@ describe('Dropdown component test suite =>', () => {
         expect(onClear).toBeCalled();
         expect(getByText('Select...')).toBeTruthy();
       });
+
+
+      it('should support custom option structure', async () => {
+        const customOptions = [
+          { firstName: 'Eve', lastName: 'Hill' },
+          { firstName: 'Justin', lastName: 'Case' },
+          { firstName: 'Anna', lastName: 'Conda' },
+        ]
+        const bindLabel = (option) => option.lastName;
+
+        const { getByText, queryByText } = render(
+          <Dropdown
+            isMultiSelect
+            options={customOptions}
+            onChange={onChange}
+            onClear={onClear}
+            bindLabel={bindLabel}
+          />
+        );
+
+        const input = screen.getByRole('textbox');
+        userEvent.click(input);
+
+        expect(getByText('Hill')).toBeInTheDocument();
+        expect(getByText('Case')).toBeInTheDocument();
+        expect(getByText('Conda')).toBeInTheDocument();
+
+
+        userEvent.click(getByText('Case'));
+        expect(getByText('Case')).toBeInTheDocument();
+        expect(queryByText('Hill')).not.toBeInTheDocument();
+        expect(queryByText('Conda')).not.toBeInTheDocument();
+      });
     });
 
     describe('when is Multiselect', () => {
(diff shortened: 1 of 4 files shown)


packages/components/src/components/dropdown/Dropdown.tsx
```diff
@@ -290,7 +292,8 @@ export class Dropdown<T = LabelValue> extends React.Component<
     menuPortalStyles: {},
     menuShouldBlockScroll: false,
     menuShouldScrollIntoView: true,
-    size: 'medium'
+    size: 'medium',
+    bindLabel: (option) => option.label,
```
symphony-annaLlorens, line 296: Why we need to have a defaultValue if it's not provided?

packages/components/spec/components/dropdown/Dropdown.spec.tsx
```diff
@@ -108,6 +108,39 @@ describe('Dropdown component test suite =>', () => {
         expect(onClear).toBeCalled();
         expect(getByText('Select...')).toBeTruthy();
       });
+
+
+      it('should support custom option structure', async () => {
```
symphony-annaLlorens, line 113: What happens if bindLabel is not provided and there are this `customOptions`?

packages/components/spec/components/dropdown/Dropdown.spec.tsx
```diff
@@ -108,6 +108,39 @@ describe('Dropdown component test suite =>', () => {
...
+        const customOptions = [
+          { firstName: 'Eve', lastName: 'Hill' },
+          { firstName: 'Justin', lastName: 'Case' },
+          { firstName: 'Anna', lastName: 'Conda' },
+        ]
+        const bindLabel = (option) => option.lastName;
+
+        const { getByText, queryByText } = render(
+          <Dropdown
```
symphony-annaLlorens, line 122: Hey, could you please move it to the `Multiselect` test case section? :)

symphony-annaLlorens:
Nice, thanks for contributing on UTK 👍

packages/components/src/components/dropdown/Dropdown.tsx
antoinerollindev: @symphony-annaLlorens If we use an option structure which is different from `{ label: '', value: '' }`, we have to provide a way to render the option text, through `bindLabel` (which actually corresponds to the `getOptionLabel` prop of the form-react-select components - along with the `getOptionValue`).

```js
<Dropdown 
    options={[
        { uid: '1', test: 'option 1'},
        { uid: '2', test: 'option 2'},
    ]} 
    bindLabel={(option) => option.text}
    bindValue={(option) => option.uid} 
/>
```

If we don't provide the `bindLabel` it will try to render the default `option.label`, and we will end up with the result of first video (see PR description).

symphony-jean-michael (approved):
LGTM

packages/components/src/components/dropdown/CustomRender.tsx
```diff
@@ -178,8 +178,8 @@ export const DropdownList = ({selectProps, ...props}: any) => {
     }
     focusThis = focusThis || searchHeaderOption;
     // Clear the value if header option is selected
-    if (selectValueSync && selectValueSync[0]?.searchHeader ) {
-        select?.clearValue();
+    if (selectValueSync && selectValueSync[0]?.searchHeader) {
```
symphony-annaLlorens, line 181: Is this for the multiselect?

symphony-annaLlorens (approved):

PR #461, opened 2022-02-27
Allow modal to accept html props

liuting1014:
## Description

The modal component has a hardcoded tabIndex "-1", there is no way to change it as it currently does not accept regular html props.  This prevents us from focusing the modal using sequential keyboard navigation. 

## JIRA ticket

[JIRA_Ticket](https://perzoinc.atlassian.net/jira/software/c/projects/RTC/boards/91?modal=detail&selectedIssue=RTC-9700&quickFilter=5206t)

### DIFF
--- a/packages/components/src/components/modal/Modal.tsx
+++ b/packages/components/src/components/modal/Modal.tsx
@@ -3,7 +3,7 @@ import * as ReactDOM from 'react-dom';
 import classNames from 'classnames';
 import { Keys } from '../common/eventUtils';
 
-type ModalProps = {
+interface ModalProps extends Omit<React.HTMLProps<HTMLDivElement>, 'size'> {
   size: 'small' | 'medium' | 'large' | 'full-width';
   className?: string;
   children?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


packages/components/src/components/modal/Modal.tsx
```diff
@@ -45,7 +45,7 @@ export const ModalFooter: React.FC<ModalContentProps> = ({
   ...rest
 }: ModalContentProps) => <div className={classNames(buildClass('footer'), className)} {...rest}> {children}</div >;
 
-const Modal: React.FC<ModalProps> = ({
+const Modal: React.FC<ModalProps> & React.HTMLProps<HTMLDivElement> = ({
```
axeleriksson147, line 48: I think `size` should be omitted as it's a custom prop on the component.

axeleriksson147 (approved):

packages/components/src/components/modal/Modal.tsx
liuting1014: Good point! I will update it !

AliciaSymphony (approved):
LGTM

symphony-adnane (approved):

PR #466, opened 2022-03-01
App-4963 table pagination

Symphony-cyril:
## Description

Adding pagination on table and css
figma for table: https://www.figma.com/file/V0VKw74594EyCKHwqYk6LV/%E2%9D%96-SDS-Core-Component-Library?node-id=5029%3A142403

## JIRA ticket

[JIRA_Ticket](https://perzoinc.atlassian.net/browse/APP-4963)

## Demo

image


packages/components/spec/components/dropdown/Dropdown.spec.tsx
```diff
@@ -108,6 +108,39 @@ describe('Dropdown component test suite =>', () => {
         expect(onClear).toBeCalled();
         expect(getByText('Select...')).toBeTruthy();
       });
+
+
+      it('should support custom option structure', async () => {
```
symphony-adnane, line 113: interesting that you had the idea of adding this test case in dropdown
What was your motivation?

packages/components/src/components/table/Table.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const [order, setOrder] = React.useState('DSC');
+  const [headingKey, setHeadingKey] = React.useState({
+    company: true,
+    contact: true,
+    country: true,
+  });
+
+  const renderIcon = (key) => {
+    if ([key] && headingKey[key]) {
+      return <Icon iconName="drop-down" onClick={() => sortingColumn(key)} />;
```
symphony-adnane, line 33: it's better to use a verb instead of a noun 
sortColumn if it executes the action of sorting ;)

packages/components/src/components/table/Table.tsx
```diff
@@ -0,0 +1,191 @@
...
+    } else {
+      return <Icon iconName="drop-up" onClick={() => sortingColumn(key)} />;
+    }
+  };
+
+  const sortingColumn = (key) => {
+    if (key && order === 'ASC') {
```
symphony-adnane, line 40: small improvement if you don't want to repeat key 
you could add a line before 
if(!key) return;
then focus only on the order

packages/components/src/components/table/Table.tsx
```diff
@@ -0,0 +1,191 @@
...
+      const sorted = [...data].sort((a, b) => (a[key] > b[key] ? 1 : -1));
+      setData(sorted);
+      setOrder('ASC');
+      setHeadingKey({ ...headingKey, [key]: false });
+    }
+  };
+
+  const handleSelectChange = (e) => {
+    setCurrentPage(1);
+    setSelect(e.target.value);
```
symphony-adnane, line 55: I think it's better if your state's name is **rowsPerPage** instead of **select**

packages/components/src/components/table/Table.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const totalPages = getTotalPages(getRowsPerPages(select), data);
+  const nextPage = hasNextPage(currentPage, getRowsPerPages(select), data);
+
+  return (
+    <>
+      <table className="tk-table">
+        <thead>
+          <tr>
+            <th className="tk-table-checkbox">
+              {showCheckbox && <Checkbox label="" name="simple-checkbox" />}
```
symphony-adnane, line 160: isn't it better if we don't generate the th element if not needed? 
the condition of showCheckbox should maybe be in line 159

packages/styles/src/components/atoms/_table.scss
```diff
@@ -1,19 +1,73 @@
...
     &:hover {
-    background-color: $--tk-table-hover-color;
+      background-color: getColor($electricity, '10');
     }
   }
   td {
     height: $--tk-table-cell-height;
     vertical-align: middle;
     border-bottom: toRem(1) solid $scolor-graphite-minus-48;
+    font-family: Segoe UI;
```
symphony-adnane, line 21: font family shouldn't be defined here

packages/styles/src/components/atoms/_table.scss
```diff
@@ -1,19 +1,73 @@
...
+    font-size: 14px;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 20px;
+    letter-spacing: 0em;
+    text-align: left;
+    color: getColor($graphite, '80');
   }
 }
+.tk-table h3 {
```
symphony-adnane, line 31: why h3?

PR #476, opened 2022-03-08
doc(checkbox): add types documentation

AliciaSymphony:
## Description

Checkbox & Radio: add types in the documentation

SelectionInput already had the props type, but not the Checkbox and Radio
This is not a breaking change, as these component were already restrited to the types (implictly from SelectionInput). I am just making it explicit

## Demo
Before
image
image

After
image
image


symphony-jean-michael (approved):
LGTM

PR #561, opened 2022-11-22
Add c9-logo icon

liuting1014:


benoit-sy (changes requested):
By convention all icons should be defined as a black path. We do not support color.
If you really want to have a Cloud9 icon with this color, you won't be able to use UITK icons.

liuting1014:
> By convention all icons should be defined as a black path. We do not support color. If you really want to have a Cloud9 icon with this color, you won't be able to use UITK icons.

Yeah, I suspected so. But thanks for confirming :) I will try to use an image directly.

PR #575, opened 2022-12-22
RTC-13661 Allow to disable Modal scrollbar

baphony:
Allow to disable the body scrollbar.

In some use cases we want to manage the scrollbar on a body child element, in that case we need to disable the body scrollbar:
image

https://perzoinc.atlassian.net/browse/RTC-13661


baphony:
Fixed there instead: https://github.com/SymphonyOSF/SFE-RTC/pull/4271

PR #591, opened 2023-03-20
Fix output to not include a src/ folder

axeleriksson147:


packages/components/tsconfig.esm.json
```diff
@@ -18,9 +18,7 @@
     "moduleResolution": "node"
   },
   "include": [
-    "src/**/*.ts",
-    "spec/**/*.ts",
-    "stories/**/*.ts"
```
axeleriksson147, line 23: Adding a TypeScript compile error in either spec/stories isn't caught. Removing them they're not part of the build and the `correct` structure in `dist` is intact.

packages/components/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
```
axeleriksson147, line 1: Is there any reason to keep two different tsconfigs? All projects should support `import` syntax even if they don't tree shake it.

symphony-jean-michael (approved):

benoit-sy (approved):

PR #594, opened 2023-03-22
Add data-testid to svg icons

axeleriksson147:
When using font icons we used to be able to target icons by their className. Adding data-testid will give us the same capabilities with svg icons.


benoit-sy (approved):

packages/components/src/components/icons/Accent.tsx
```diff
@@ -1,7 +1,13 @@
 import * as React from 'react';
 import { SVGProps } from 'react';
 const SvgAccent = (props: SVGProps<SVGSVGElement>) => (
-  <svg width="1em" height="1em" viewBox="0 0 16 16" {...props}>
+  <svg
```
axeleriksson147, line 4: All these React components are auto-generated running `svgr`.

symphony-jean-michael (approved):

PR #608, opened 2023-10-14
Switch from classNames to clsx

axeleriksson147:
`clsx` is a modern version of `classNames`. `classNames` only exports CommonJS and causes some issues for using certain libraries in `SFE-RTC`.

https://areknawo.com/you-dont-know-the-classnames-library/


axeleriksson147:
Seems like our CircleCI integration is gone?

swasunb (approved):

antoinerollindev (approved):

PR #609, opened 2023-10-16
Replace lodash with native methods

axeleriksson147:
We have the same problem here that `lodash` isn't exporting a default property. `lodash-es` does but it's not compatible with `enzyme` and `jest` that is used for unit tests. Trying to upgrade `jest` to have ESM compatibility, it needs a later version of Node than 12. The errors just cascade. I can fix this in a subsequent PR but would like to get the packages working in SFE-RTC and for that we cannot have older CommonJS modules that don't have a default export.

What would be sensible would be to change to @testing-library that is what's standard for component unit testing.

Also added a "yarn dedupe" command and ran it. The previous `yarn.lock` was quite bloated.


packages/components/src/core/validators/validators.ts
```diff
@@ -1,4 +1,4 @@
-import * as _ from 'lodash';
+const isEmpty = obj => [Object, Array].includes((obj || {}).constructor) && !Object.entries((obj || {})).length;
```
benoit-sy, line 1: Could you please put that in a utility file instead ?

packages/components/src/utils/debounce.ts
benoit-sy, line 1: Couldn't we found a smaller implementation of debounce? Sounds a bit big here to pull all that

benoit-sy (changes requested):

packages/components/src/utils/debounce.ts
axeleriksson147: Yes, we have one in `SFE-RTC` but it doesn't work exactly the same, nor does the others from for example https://youmightnotneed.com/lodash

How about this is temporary, and I will subsequently fix the underlying issues, then we can put back the package and get rid of this massive utility file.

packages/components/src/core/validators/validators.ts
axeleriksson147: Updated!

benoit-sy (approved):

antoinerollindev (approved):

PR #610, opened 2023-10-17
Use better-prop-types as it exports ESM

axeleriksson147:
Last hurdle (for now) to get things working in SFE-RTC. `better-prop-types` exports as ESM.

Personally I don't think we should use `prop-types` at all. It's unnecessary as it catches issues at runtime and VSCode and other IDE integrations help with types even if you're writing JS.


antoinerollindev (approved):

packages/components/package.json
```diff
@@ -36,11 +36,12 @@
     "@popperjs/core": "^2.4.4",
     "@seznam/compose-react-refs": "^1.0.6",
     "@symphony-ui/uitoolkit-styles": "3.6.10",
+    "better-prop-types": "^1.1.1",
     "clsx": "^2.0.0",
     "date-fns": "^2.16.1",
     "emoji-js": "^3.5.0",
     "github-markdown-css": "^4.0.0",
-    "prop-types": "^15.7.2",
+    "prop-types": "^15.8.1",
```
swasunb, line 44: Is this still needed?

axeleriksson147: Yes, as I didn't change the other components. Seems like `better-prop-types` has the same issue with the unit test runner and being an ESM module. Might have to upgrade it...

PR #614, opened 2023-10-18
Update Jest to latest

axeleriksson147:
- Revert the tsconfig changes as it didn't work for some custom Storybook being run in the extension team.
- Update to Jest latest to allow test runner to run ESM
- Add open-ssl-legacy-provider to commands


packages/components/tsconfig.esm.json
```diff
@@ -0,0 +1,29 @@
+{
```
swasunb, line 1: If we are to keep both tsconf, it would be better to have a base one and extend only with the necessary ones that differs between esm and cjs

swasunb (changes requested):

packages/components/tsconfig.esm.json
axeleriksson147: Yeah, can do that.

swasunb (approved):

packages/components/tsconfig.esm.json
axeleriksson147: But it's kind of pointless as compilerOptions doesn't get merged but overwritten so would have to duplicate all of them.

axeleriksson147: So leaving as is.

antoinerollindev (approved):

PR #615, opened 2023-10-19
Upgrade to latest Storybook (7)

axeleriksson147:
Sorry the change size is huge but I think it's pretty straight forward going through each file. All the stories converted follow the same pattern.

Upgrade to Storybook 7. It was a lot more changes than I anticipated because all stories had to be fixed (a lot of stories were never working in the first place). But this should have no affect on the output nor any other packages running their own Storybook on the components, such as `eis-common-ui`.

The only thing dropped is the dark/light theme but I will fix that in a subsequent PR. I think the scope of this one is enough.


.yarnrc
```diff
@@ -0,0 +1 @@
+registry "https://registry.yarnpkg.com"
```
axeleriksson147, line 1: As this is a public package, when adding dependencies we should assure we use the public repository.

packages/components/.storybook/main.js
```diff
@@ -1,16 +1,25 @@
...
+    getAbsolutePath('@storybook/addon-docs'),
+    getAbsolutePath('@storybook/addon-actions'),
+    getAbsolutePath('@storybook/addon-links'),
+    getAbsolutePath('@storybook/addon-a11y'),
+    getAbsolutePath('storybook-addon-mdx-embed'),
+    getAbsolutePath('storybook-dark-mode'),
+    getAbsolutePath('@storybook/addon-controls'),
+    getAbsolutePath('@storybook/preset-scss'),
+    getAbsolutePath('storybook-addon-themes'),
   ],
```
axeleriksson147, line 14: Added by the Storybook migration tool.

packages/components/package.json
```diff
@@ -19,9 +19,9 @@
     "svg-icons-to-react-components": "svgr --icon --out-dir src/components/icons -- ../styles/src/icons/svg",
     "lint": "eslint --ext .ts,.tsx,.js,.jsx ./",
     "lint-fix": "eslint --ext .ts,.tsx,.js,.jsx ./ --fix",
-    "start": "NODE_OPTIONS=--openssl-legacy-provider start-storybook -p 6006",
```
axeleriksson147, line 22: Not needed anymore.

packages/components/stories/Avatar.stories.tsx
```diff
@@ -1,372 +1,389 @@
+import './styles/avatar.stories.css';
```
axeleriksson147, line 1: All stories work as before or have been fixed because they were never rendered.

packages/styles/.babelrc.json
```diff
@@ -0,0 +1,16 @@
+{
```
axeleriksson147, line 1: Created by Storybook 7 migration helper.

packages/components/spec/components/validation/Validation.spec.tsx
```diff
@@ -1,11 +1,11 @@
...
-import { DatePicker, TextField, Validation } from '../../../src/components';
+import { TextField, Validation } from '../../../src/components';
 import { Validators } from '../../../src/core/validators/validators';
 
-function currentEventLoopEnd() {
-  return new Promise(resolve => setImmediate(resolve));
-}
+// function currentEventLoopEnd() {
+//   return new Promise(resolve => setImmediate(resolve));
+// }
```
antoinerollindev, line 8: Might need to remove this?

antoinerollindev (approved):

PR #616, opened 2023-10-23
Update ESLint

axeleriksson147:


axeleriksson147:
Merging as it's an ESLint update.

PR #621, opened 2023-12-11
RTC-14706 - Switch Dropdown icons to be SVGIcons

axeleriksson147:
As previously discussed font icon rendering started breaking at times since Startpage was introduced and the solution has been to switch to SVGIcons which is a more modern way of doing it. This fixes the `Dropdown` component.

### DIFF
--- a/packages/components/spec/components/time-picker/TimePicker.spec.tsx
+++ b/packages/components/spec/components/time-picker/TimePicker.spec.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
-import { mount, shallow } from 'enzyme';
+import { mount } from 'enzyme';
 import { act } from 'react-dom/test-utils';
 
 import TimePicker from '../../../src/components/time-picker/TimePicker';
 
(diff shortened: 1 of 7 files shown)


swasunb (approved):

packages/components/spec/components/time-picker/TimePicker.spec.tsx
```diff
@@ -39,25 +38,15 @@ describe('TimePicker Component', () => {
...
-    expect(wrapperPicker.prop('onCut')).toBe(props.onCut);
-    expect(wrapperPicker.prop('onDrag')).toBe(props.onDrag);
+  it('should render', () => {
+    render(<TimePicker />);
   });
 
+  it('should properly pass props to the underlying Dropdown component', () => {
+    render(<TimePicker label="New Label"/>);
+    screen.getByText('New Label');
+  })
```
axeleriksson147, line 48: Using React Testing Library instead of Enzyme.

PR #628, opened 2024-04-09
Refactor the Tooltip component and use modern @floating-ui/react library instead of react-popper

axeleriksson147:
This PR refactors the Tooltip component and uses the sequel to `react-popper`, `@floating-ui/react` - https://github.com/floating-ui/floating-ui.

- The code is easier to understand.
- Less libraries involved.
- Easier to do component testing.

All tests pass and I've tested and compared the components manually against the old Storybook and the component behaves the same or better.


packages/components/package.json
```diff
@@ -36,19 +36,18 @@
     "node": ">=14.21.2"
   },
   "dependencies": {
+    "@emotion/cache": "^11.11.0",
+    "@emotion/react": "^11.11.4",
```
axeleriksson147, line 40: These dependencies were used in the `StylesInjection.tsx` but not explicitly added as dependencies.

packages/styles/src/components/atoms/_hint-or-tooltip.scss
```diff
@@ -1,41 +0,0 @@
-.tk-hint-or-tooltip {
```
axeleriksson147, line 1: Provided by `@floating-ui/react`.

packages/components/src/components/tooltip/Tooltip.tsx
```diff
@@ -65,140 +53,123 @@ const Tooltip: React.FC<TooltipProps> = ({
...
-  return (
-    <>
-      { /**
-       * The element tooltip is wrapped around.
-       */ }
-      { /** Extra span to make sure Tooltip works on disabled children */ }
-      {displayTrigger === 'hover' &&
-          showTooltipOnHover(children, handleMouseMove)}
-      {displayTrigger === 'click' &&
-          showTooltipOnClick(children, showClick, setShowClick)}
```
axeleriksson147, line 126: Before clicking anywhere on a hint would close it if displayTrigger was `click` which is incorrect behaviour. Now it'll only close if close label is clicked, the reference element or if you click outside.

packages/components/src/components/tooltip/Tooltip.tsx
```diff
@@ -65,140 +53,123 @@ const Tooltip: React.FC<TooltipProps> = ({
...
   type: 'hint',
 };
 
 Tooltip.propTypes = {
+  className: PropTypes.string,
   children: PropTypes.any,
   closeLabel: PropTypes.string,
   description: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
   displayTrigger: PropTypes.oneOf(['click', 'hover']),
+  hoverDelay: PropTypes.number,
```
antoinerollindev, line 168: Don't we miss hoverTimeout here? 🤔

antoinerollindev (approved):

PR #631, opened 2024-07-24
Fixing tooltip component

axeleriksson147:
Tooltip doesn't disappear when hovering when a new DOM element overlays the previous one. Using an updated version of @floating-ui/react I can't reproduce it.

### DIFF
--- a/packages/components/src/components/tooltip/Tooltip.tsx
+++ b/packages/components/src/components/tooltip/Tooltip.tsx
@@ -1,4 +1,3 @@
-import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import { clsx } from 'clsx';
 import {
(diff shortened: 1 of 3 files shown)


packages/components/src/components/tooltip/Tooltip.tsx
```diff
@@ -1,176 +1,155 @@
...
-    </div>
-    }
-  </>
-};
-
-Tooltip.defaultProps = {
-  type: 'hint',
-};
-
-Tooltip.propTypes = {
```
axeleriksson147, line 162: @antoinerollindev @aurelie-forzani @benoit-sy do you get usage from prop-styles package? I feel like TypeScript should be enough and it's a burden having both maintained.

benoit-sy: @axeleriksson147 The only use-case where PropTypes still makes sense over pure TypeScript is for when dynamic data are passed to components.
So we could maybe plan for a global refactor PR where we ditch entirely prop-types in favour or TS types/interfaces (that most of the time we already have)

axeleriksson147: Yeah, and in that case it's about getting an "easier" error message but the component will error either way I believe.

benoit-sy (approved):
✅ for a `v4.2.1`

PR #633, opened 2024-08-21
BYC-1304 Handle focus trap on the Modal component

symphony-jean-michael:
## Description

Manage the trap of focusing on the modal component.
The added logic is quite simple, and solves simple scenarios. I'll add complexity later if more complex situations come up.

This behavior is set by default, because this is the default behavior recommended by accessibility guidelines. 
>When a modal element is displayed, focus should be placed in the modal. Focus needs to be "trapped" inside the modal when it is visible, until it is dismissed.
(Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal)

However, it can be deactivated with `focusTrapEnabled={false}`.

## JIRA ticket

[JIRA_Ticket](https://perzoinc.atlassian.net/browse/BYC-1304)

## Demo

https://github.com/user-attachments/assets/a534b614-9bc6-4f53-990c-505441f273d2

### DIFF
--- a/packages/components/spec/components/modal/Modal.spec.tsx
+++ b/packages/components/spec/components/modal/Modal.spec.tsx
@@ -140,7 +140,7 @@ describe('Modal', () => {
   describe('should handle focus', () => {
     const component = <div>
       <a href="#">Another element that should not be on focus</a>
-      <Modal focusTrapEnabled size={'small'} show={true}>
+      <Modal size={'small'} show={true}>
         <ModalHeader>Hello, World</ModalHeader>
         <ModalBody>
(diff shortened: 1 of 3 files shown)


ZimmySymphony (approved):

antoinerollindev (approved):
👍 
Should we bump the package version numbers? (can't remember the release process)

axeleriksson147 (approved):

antoinerollindev:
Do you think we could have cases where we prefer the previous accessibility behaviour ?
If not, we can directly set it as default behaviour 🤔

tzhao-symphony (approved):

PR #646, opened 2024-09-11
Banner: close button reset

benoit-sy:
Before (broken)

image

After
image

### DIFF
--- a/packages/styles/src/components/atoms/_banner.scss
+++ b/packages/styles/src/components/atoms/_banner.scss
@@ -9,6 +9,11 @@
         margin-right: toRem(8);
     }
 
+    &__close {
+        background: unset;
+        border: unset;
+    }
+
     &__content {
         flex: 1;
         text-align: left;


antoinerollindev (approved):
The screenshot in the description is the "before" or the "after" ?
If it's the new design, we could probably improve the styling of the close button ?

swasunb (approved):

benoit-sy:
the component itself is following SDS. here the usage is a TOP banner like the one we have in Client 2

philt-symphony:
> the component itself is following SDS. here the usage is a TOP banner like the one we have in Client 2

I was talking about the cross, it does not look at all as in the SDS :)

benoit-sy:
> The screenshot in the description is the "before" or the "after" ?

@antoinerollindev the screenshot shows the issue, which is the broken button 😄 . There is "somewhere" in UITK a piece of code that reset button, but when using only the banner, it's broken ... so hence the PR

antoinerollindev:
> > The screenshot in the description is the "before" or the "after" ?
> 
> @antoinerollindev the screenshot shows the issue, which is the broken button 😄 . There is "somewhere" in UITK a piece of code that reset button, but when using only the banner, it's broken ... so hence the PR

Perfect ! Looks great 👍

philt-symphony (approved):

PR #664, opened 2024-12-18
[CES-20277] Input in dropdown should be of type search

aurelie-forzani:
## Description

Input in dropdown should be of type search
## JIRA ticket

[JIRA_Ticket](https://perzoinc.atlassian.net/browse/CES-20277)


aurelie-forzani:
ok to test

aurelie-forzani:
ok to test

PR #670, opened 2025-01-09
C2-24526-uitoolkit: [WCAG] [Calendar picker] enhance accessibility

NamNguyenFPT:
## Description
- Move code from Lite to ui-toolkit
- Enhance accessibility

## JIRA ticket
[C2-24526](https://perzoinc.atlassian.net/browse/C2-24526)

## Demo

Put screenshots, videos or gif demos of what is the behaviour after your PR.

### DIFF
--- a/packages/components/src/components/date-picker/DatePicker.tsx
+++ b/packages/components/src/components/date-picker/DatePicker.tsx
@@ -53,6 +53,8 @@ type DatePickerComponentProps = {
   onCalendarOpen?: (datePickerID: string) => void;
   /** Handle calendar close event */
   onCalendarClose?: (datePickerID: string) => void;
+  clearData?: boolean;
+  datePickerID?: string;
   placeholder?: string;
   locale?: string;
   placement?: 'top' | 'bottom' | 'right' | 'left';
@@ -190,10 +192,10 @@ class DatePicker extends Component<
   componentDidUpdate(prevProps, prevState) {
     if (this.state.showPicker && !prevState.showPicker) {
       this.mountDayPickerInstance();
-      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.id);
+      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.datePickerID);
     } else if (!this.state.showPicker && prevState.showPicker) {
       this.unmountDayPickerInstance();
-      this.props.onCalendarClose && this.props.onCalendarClose(this.props.id);
+      this.props.onCalendarClose && this.props.onCalendarClose(this.props.datePickerID);
       if(this.props.shouldResetInvalidDate) {
         const validDate = this.props.date && this.props.date;
         this.setState({
@@ -219,6 +221,11 @@ class DatePicker extends Component<
             locale: getLocale,
           })
           : null,
+      });
+    }
+    if (this.props.clearData !== prevProps.clearData && this.props.clearData) {
+      this.setState({
+        inputValue: null,
       });
     }
   }
@@ -445,7 +452,7 @@ class DatePicker extends Component<
       break;
     case Keys.ESC:
       cancelEvent(e);
-      this.handleOnClose();
+      this.handleOnEscAction(e);
       break;
     default:
       break;
@@ -484,6 +491,16 @@ class DatePicker extends Component<
     }
   }
 
+  private handleOnEscAction(e) {
+    const { showPicker } = this.state;
+    if (showPicker) {
+      this.handleOnClose();
+    } else {
+      const element = (e.target as HTMLElement).parentNode;
+      element?.dispatchEvent(new KeyboardEvent(e.type, e));
+    }
+  }
+
   renderCalendar() {
     const {
       date,
@@ -494,6 +511,7 @@ class DatePicker extends Component<
       labels,
       todayButton,
       ariaLabel,
+      datePickerID,
       menuPortalTarget,
       menuPortalStyles,
     } = this.props;
@@ -523,6 +541,7 @@ class DatePicker extends Component<
           month={navigationDate}
           todayButton={todayButton}
           labels={labels}
+          datePickerID={datePickerID}
           onDayClick={this.handleDayClick}
           onClose={this.handleOnClose}
         />
--- a/packages/components/src/components/date-picker/sub-component/DayPicker.tsx
+++ b/packages/components/src/components/date-picker/sub-component/DayPicker.tsx
@@ -53,6 +53,7 @@ type DayPickerComponentProps = {
   month?: Date;
   selectedDays?: Date;
   todayButton?: string;
+  datePickerID?: string;
   onClose?: () => any;
   onDayClick?: (date: Date, modifiers) => any;
 };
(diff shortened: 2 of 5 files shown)


tri-nguyenduc (approved):
LGTM

phuocdinh-sym (approved):

packages/components/src/components/date-picker/sub-component/DayPicker.tsx
```diff
@@ -458,6 +461,7 @@ class DayPicker extends React.Component<
         className="tk-daypicker"
         ref={(el) => (this.dayPicker = el)}
         onKeyDown={this.handleKeyDownContainer}
+        id={datePickerID}
```
symphony-jean-michael, line 464:
```suggestion
        id={id}
```

packages/components/src/components/date-picker/sub-component/DayPicker.tsx
```diff
@@ -449,7 +452,7 @@ class DayPicker extends React.Component<
   }
 
   render() {
-    const { dir, labels, locale } = this.props;
+    const { dir, labels, locale, datePickerID } = this.props;
```
symphony-jean-michael, line 455:
```suggestion
    const { dir, labels, locale, id } = this.props;
```

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -190,10 +192,10 @@ class DatePicker extends Component<
   componentDidUpdate(prevProps, prevState) {
     if (this.state.showPicker && !prevState.showPicker) {
       this.mountDayPickerInstance();
-      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.id);
+      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.datePickerID);
```
symphony-jean-michael, line 195: Why don't you use the `id` ?

NamNguyenFPT: Hi @symphony-jean-michael, ID is used for input so I want to use another variable to use for DatePicker. This makes it easy for us to write automation tests for it.

symphony-jean-michael: Ah ! Yes, good catch !

symphony-jean-michael (approved):

packages/components/src/components/date-picker/sub-component/DayPicker.tsx
```diff
@@ -53,6 +53,7 @@ type DayPickerComponentProps = {
   month?: Date;
   selectedDays?: Date;
   todayButton?: string;
+  datePickerID?: string;
```
benoit-sy, line 56: At the `<DayPicker>` level the prop should be named `id` and not `datePickerID`

packages/components/src/components/date-picker/sub-component/DayPicker.tsx
benoit-sy, line 464:
```suggestion
        id={id}
```

packages/components/src/components/date-picker/sub-component/DayPicker.tsx
benoit-sy, line 455:
```suggestion
    const { dir, labels, locale, id } = this.props;
```

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -53,6 +53,8 @@ type DatePickerComponentProps = {
   onCalendarOpen?: (datePickerID: string) => void;
   /** Handle calendar close event */
   onCalendarClose?: (datePickerID: string) => void;
+  clearData?: boolean;
+  datePickerID?: string;
```
benoit-sy, line 57:
```suggestion
  dayPickerId?: string;
```

benoit-sy: it's applied on the`<DayPicker>` it's very confusing to name it DatePicker

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -190,10 +192,10 @@ class DatePicker extends Component<
   componentDidUpdate(prevProps, prevState) {
     if (this.state.showPicker && !prevState.showPicker) {
       this.mountDayPickerInstance();
-      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.id);
+      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.datePickerID);
     } else if (!this.state.showPicker && prevState.showPicker) {
       this.unmountDayPickerInstance();
-      this.props.onCalendarClose && this.props.onCalendarClose(this.props.id);
+      this.props.onCalendarClose && this.props.onCalendarClose(this.props.datePickerID);
```
benoit-sy, line 198: Please add your new `dayPickerId` as a second  optional argument because otherwise it's a **breaking change**

benoit-sy (changes requested):
Please have a look at my comments. I am not convinced of your changes on the Datepicker overalll.

If the change is for only testing purpose, please add a `data-testid` instead and don't mess around with public props

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -50,9 +50,11 @@ type DatePickerComponentProps = {
   /** Handle focus event */
   onFocus?: (event: React.FocusEvent<HTMLElement>) => void;
   /** Handle calendar open event */
-  onCalendarOpen?: (datePickerID: string) => void;
+  onCalendarOpen?: (dayPickerId: string) => void;
```
benoit-sy, line 53: You're changing a public api. this is a breaking change

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -50,9 +50,11 @@ type DatePickerComponentProps = {
...
   /** Handle calendar close event */
-  onCalendarClose?: (datePickerID: string) => void;
+  onCalendarClose?: (dayPickerId: string) => void;
```
benoit-sy, line 55: You're changing a public api. this is a breaking change

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -484,6 +491,16 @@ class DatePicker extends Component<
...
+  private handleOnEscAction(e) {
+    const { showPicker } = this.state;
+    if (showPicker) {
+      this.handleOnClose();
+    } else {
+      const element = (e.target as HTMLElement).parentNode;
+      element?.dispatchEvent(new KeyboardEvent(e.type, e));
+    }
+  }
+
```
benoit-sy, line 503: Could you explain the rationale behind this change ?

benoit-sy (changes requested):
The only change/update I understand is the `div` to `button` transformation with proper roles and aria attributes.

For the other one, I'm struggling trying to understand them looking at the code.

packages/components/src/components/date-picker/DatePicker.tsx
NamNguyenFPT: After pressing Esc from the keyboard, if the datePicker is open, it'll close it. Otherwise, it triggers the Esc event and assigns it to the parrent element =>that is Advanced Search Modal. Refer to this bug https://perzoinc.atlassian.net/browse/C2-24820 => this is the reason to have above code

packages/components/src/components/date-picker/DatePicker.tsx
NamNguyenFPT: Hi @benoit-sy, The functions onCalendarOpen and onCalendarClose are functions I added to handle closing and opening DatePicker, so declaring them means you have to add all the parameters they require => this is not a breaking change

benoit-sy: It is a breaking change in the sense you did not added them in this same PR. they have been there with a specific signature, made available in a public build, so people could have started using them, and will wait for the first argument to be the datepicker id.

packages/components/src/components/date-picker/DatePicker.tsx
benoit-sy: I see your point, but this is not how you should fix it.

```suggestion
  private handleOnEscAction(e) {
    const { showPicker } = this.state;
    if (showPicker) {
      this.handleOnClose();
      e.stopPropagation();
    }
  }

```

benoit-sy (changes requested):
I am sorry to insist, but I still don't understand why you need this new daypicker id and why you need it as argument of the `onCalendarOpen` and `onCalendarClose` callbacks

Again if it is just a matter of testing purpose, please add generated data-testid

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -190,10 +192,10 @@ class DatePicker extends Component<
   componentDidUpdate(prevProps, prevState) {
     if (this.state.showPicker && !prevState.showPicker) {
       this.mountDayPickerInstance();
-      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.id);
+      this.props.onCalendarOpen && this.props.onCalendarOpen(this.props.dayPickerId);
```
benoit-sy, line 195: this is the actual breaking change I am talking about above

packages/components/src/components/date-picker/DatePicker.tsx
NamNguyenFPT: Hi @benoit-sy, I've already worked around the Esc function in Lite; it would be preferable if we moved it to ui-toolkit so that we wouldn't have to do it again if DatePicker is called in another location later.

benoit-sy: @NamNguyenFPT my point was that the fix/workaround in Lite should be removed as not needed at all, and the only fix to be done (that would fix the Lite issue we had 3 months ago) is my suggestion above.

when we hit ESC in the datepicker, and it is opened, we should close it and stop the propagation of the event, as it is already consumed by the datepicker, and we don't want any parent node to consume that event as it will bubble up the dom tree.

benoit-sy (changes requested):
Please have a look at my review comment. This is the last one remaining, all other changes are good and ✅

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -484,6 +490,16 @@ class DatePicker extends Component<
...
+  private handleOnEscAction(e) {
+    const { showPicker } = this.state;
+    if (showPicker) {
+      this.handleOnClose();
+    } else {
+      const element = (e.target as HTMLElement).parentNode;
+      element?.dispatchEvent(new KeyboardEvent(e.type, e));
+    }
+  }
+
```
benoit-sy, line 502: @NamNguyenFPT my point was that the fix/workaround in Lite should be removed as not needed at all, and the only fix to be done (that would fix the Lite issue we had 3 months ago) is my suggestion above.

when we hit ESC in the datepicker, and it is opened, we should close it and stop the propagation of the event, as it is already consumed by the datepicker, and we don't want any parent node to consume that event as it will bubble up the dom tree.

```suggestion
  private handleOnEscAction(e) {
    const { showPicker } = this.state;
    if (showPicker) {
      this.handleOnClose();
      e.stopPropagation();
    }
  }

```

benoit-sy (approved):

symphony-jean-michael:
ok to test

PR #678, opened 2025-02-10
C2-24526-uitoolkit: update behavior to DatePicker

NamNguyenFPT:
## Description
- add more func: focus to selected date when opening the DatePicker

## JIRA ticket
[C2-24526](https://perzoinc.atlassian.net/browse/C2-24526)

## DEMO

https://github.com/user-attachments/assets/abbea188-ca2f-462c-b806-9f56fc68f00c

[C2-24526]: https://perzoinc.atlassian.net/browse/C2-24526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/spec/components/date-picker/DatePicker.spec.tsx
+++ b/packages/components/spec/components/date-picker/DatePicker.spec.tsx
@@ -213,8 +213,27 @@ describe('DatePicker Component', () => {
       });
     });
   });
-  describe('should trigger onCalendarOpen', () => {
-    it('when opening the date picker', () => {
+  describe('when opening the date picker', () => {
+    it('should focus to current date if the value is null ', () => {
+      const currentDate = new Date().getDate();
+      const props = createTestProps({ showOverlay: true, date: null });
+      render(<DatePicker {...props} />);
+      const icon = document.querySelector('.tk-icon-calendar');
+      fireEvent.keyDown(icon, { key: Keys.ENTER });
+
+      const focusedCell = screen.getByText(`${currentDate}`);
+      expect(document.activeElement).toEqual(focusedCell);
+    });
+    it('should focus to selected date if the value is not null', () => {
+      const props = createTestProps({ showOverlay: true });
+      render(<DatePicker {...props} />);
+      const icon = document.querySelector('.tk-icon-calendar');
+      fireEvent.keyDown(icon, { key: Keys.ENTER });
+
+      const focusedCell = screen.getByText(`${props.date.getDate()}`);
+      expect(document.activeElement).toEqual(focusedCell);
+    });
+    it('should trigger onCalendarOpen', () => {
       const props = createTestProps({});
       const wrapper = shallow(<DatePicker {...props} />);
       wrapper.setState({ showPicker: true });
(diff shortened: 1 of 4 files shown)


tri-nguyenduc (approved):

packages/components/spec/components/date-picker/DatePicker.spec.tsx
```diff
@@ -222,6 +222,27 @@ describe('DatePicker Component', () => {
       expect(props.onCalendarOpen).toHaveBeenCalledTimes(1);
     });
   });
+  describe('should focus to', () => {
```
phuocdinh-sym, line 225: For each it/describe should be completed statement.
So we should describe('When opening the date picker') and it('should focus to ...')

packages/components/src/components/date-picker/utils/dateCons.ts
benoit-sy, line 1: Could you rename the file to `dateConstants.ts`

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -370,6 +373,16 @@ class DatePicker extends Component<
...
+  private handleFocusToSelectedDate() {
+    const selectedDate = this.props.date ? this.props.date.getDate() : new Date().getDate();
+    if (this.refPicker && this.refPicker.dayPicker) {
+      const dayNodes = this.refPicker.dayPicker.querySelectorAll(
+        DAYS_VISIBLE_SELECTOR
+      );
+      dayNodes[selectedDate - 1].focus();
+    }
+  }
+
```
benoit-sy, line 385: Shouldn't we do that only when the datePicker is opened ?

benoit-sy (changes requested):
Please have a look at the couple comments below

packages/components/src/components/date-picker/DatePicker.tsx
NamNguyenFPT: When this function was called, it checked that condition. It is only called once the DatePicker is opened.

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -370,6 +373,16 @@ class DatePicker extends Component<
     }
   }
 
+  private handleFocusToSelectedDate() {
+    const selectedDate = this.props.date ? this.props.date.getDate() : new Date().getDate();
```
symphony-jean-michael, line 377: `new Date()` should only be called one time in the entire file, otherwise we risk having different values in 2 variables. 🤔 
I suggest you using `state.navigationDate` instead of `props.date`. It should word. Can you try and update the PR if it works correctly ? Thanks

NamNguyenFPT: yes of course, let me try

symphony-jean-michael (commented):
Please provide screenshots / videos from the Storybook.

symphony-jean-michael (approved):

PR #681, opened 2025-02-13
[CES-20765] Fix dropdown style issue using input type as 'search'

mgeorges-symphony:
## Description
Fix a UI issue on dropdown component using an input with type="search".
 

## JIRA ticket

[JIRA_Ticket](https://perzoinc.atlassian.net/browse/CES-20765)

## Demo

https://github.com/user-attachments/assets/313e60aa-4ffa-4e58-9159-9be84ed4984b

### DIFF
--- a/packages/styles/src/uitoolkit-components/_select.scss
+++ b/packages/styles/src/uitoolkit-components/_select.scss
@@ -176,6 +176,9 @@ body {
       .tk-select__value-container {
         cursor: text;
         padding:0;
+        input[type = 'search']::-webkit-search-cancel-button {
+          display: none;
+        }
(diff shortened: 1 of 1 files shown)


antoinerollindev (approved):

symphony-jean-michael (approved):
Hi ! 
I'm not a big fan to use of using a [non-standard selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button) but it seems there's no other way. 😕
This selector is not supported on Firefox, but it seems we're fine since the bug doesn't occur on Firefox.

benoit-sy (approved):

PR #688, opened 2025-03-18
[CES-20338][Accessibility] Add outline color + change size of the close modal button on dialog component

mgeorges-symphony:
## Description
For accessibility, update of the outline color + change of the size of the focus on close modal button for the dialog component. 

## JIRA ticket

[CES-20338](https://perzoinc.atlassian.net/browse/CES-20338)

## Demo
Screenshot 2025-03-17 at 22 17 33

### DIFF
--- a/packages/styles/src/components/atoms/_dialog.scss
+++ b/packages/styles/src/components/atoms/_dialog.scss
@@ -70,12 +70,17 @@ $dialog-scrollbar-box-shadow: rgba($scolor-black, 0.5);
     right: 2rem;
     top: 2rem;
     z-index: 200;
+    height: toRem(24);
+    width: toRem(24);
 
     &>svg {
       fill: var(--tk-dialog-cross-color, #7c7f86);
 
       &:hover {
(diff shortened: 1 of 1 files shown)


packages/styles/src/components/atoms/_dialog.scss
```diff
@@ -70,13 +70,18 @@ $dialog-scrollbar-box-shadow: rgba($scolor-black, 0.5);
...
 
     &>svg {
       fill: var(--tk-dialog-cross-color, #7c7f86);
 
       &:hover {
         fill: var(--tk-dialog-hover-cross-color, #525760);
       }
+      &:focus {
+        outline-color: getColor($electricity, '40');
+      }
```
philt-symphony, line 84: I think it should be focu-visible here, or the outline will be displayed even using the mouse (and it should be displayed only using keyboard)

mgeorges-symphony: I updated the code, thanks

benoit-sy (approved):

philt-symphony (approved):

PR #698, opened 2025-04-01
C2-26005: fix failed UT

NamNguyenFPT:
## Description
- add more condition to check if dateNode has value

## JIRA ticket

[C2-26005](https://perzoinc.atlassian.net/browse/C2-26005)

## Demo
BUG

https://github.com/user-attachments/assets/c03cda66-bebd-4a33-b1d1-20140999d3ca

FIX

https://github.com/user-attachments/assets/d0c88908-19c2-4199-917d-cf7e3cd3f749

[C2-26005]: https://perzoinc.atlassian.net/browse/C2-26005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -374,9 +374,11 @@ class DatePicker extends Component<
         DAYS_VISIBLE_SELECTOR
       );
       const dateNode = dayNodes[selectedDate - 1];
-      if (dateNode.tabIndex === 0) {
+      // In a special case from UT, dateNode can be undefined,
+      // hence there is a condition to check if it has value.
+      if (dateNode && dateNode.tabIndex === 0) {
```
symphony-jean-michael, line 379:
```suggestion
      if (dateNode?.tabIndex === 0) {
```

packages/components/src/components/date-picker/DatePicker.tsx
```diff
@@ -374,9 +374,11 @@ class DatePicker extends Component<
         DAYS_VISIBLE_SELECTOR
       );
       const dateNode = dayNodes[selectedDate - 1];
-      if (dateNode.tabIndex === 0) {
+      // In a special case from UT, dateNode can be undefined,
```
symphony-jean-michael, line 377: We don't need these comments

symphony-jean-michael (commented):
IMO, the root cause of the problem is not fixed. Why dateNode is `undefined` ? I guess it's because the current date and the default select date are not the same (See description in ticket https://perzoinc.atlassian.net/browse/C2-26005 )

Screenshot 2025-03-31 at 17 57 58

Screenshot 2025-04-01 at 17 56 25

symphony-jean-michael (approved):

PR #702, opened 2025-04-10
RTC-15411 - click & keyup listener should be attached to the document the DropdownMenu is rendered in

axeleriksson147:
https://perzoinc.atlassian.net/browse/RTC-15411

We use `DropdownMenu` in SFE-RTC and if the content is in a popped out window, it will not close when clicking outside due to the reference being the main document.

This solution with using the `ref` and accessing the `ownerDocument` is how we've done it for specific SFE-RTC components.

Tested against SFE-Lite running locally and it works as intended.


packages/components/src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -1,19 +1,15 @@
...
 interface DropdownMenuProps extends React.HTMLProps<HTMLDivElement> {
   show?: boolean;
-  children?: React.ReactNode;
-  className?: string;
   onClose?: () => void;
 }
 
 interface DropdownMenuItemProps extends React.HTMLProps<HTMLDivElement> {
-  children?: React.ReactNode;
-  className?: string;
```
benoit-sy, line 16: why did you remove this from the interface ? they both are still used is each component.

axeleriksson147: Ok @benoit-sy ?

baphony (approved):

antoinerollindev (approved):
Can we add a unit test for that?

swasunb (approved):

packages/components/src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -92,14 +89,14 @@ export const DropdownMenu: React.FC<DropdownMenuProps> = ({children, className,
...
-    if (onClose && menu && !(e.composedPath() as any).includes(menu.current)) {
+    if (onClose && ref && !(e.composedPath() as any).includes(ref.current)) {
       onClose();
     }
   }
 
   const removeListeners = useCallback(() => {
-    window.removeEventListener('keyup', keyboardEventHandler);
-    window.removeEventListener('click', mouseEventHandler);
+    ref.current?.ownerDocument.removeEventListener('keyup', keyboardEventHandler);
```
axeleriksson147, line 98: The `ref.current?` is added because current can be undefined but the React 16 types don't reflect that. Another good reason to upgrade to React 17.

PR #706, opened 2025-04-15
Replace `nanoid` by `uuid`

antoinerollindev:
**Replace `nanoid` by `uuid`**

`nanoid` is causing unit test failures in projects using ui-toolkit.

- remove `nanoid` dependency
- add `uuid` dependency (no snyk issue)
- replace `nanoid()` by `uuidv4()`
- replace `customAlphabet()` by a custom function using `uuidv4()`


axeleriksson147:
Sounds good, but seems like uuid also uses the `crypto` library so I assume the same issues will arise. Let's see..

PR #715, opened 2025-04-22
RTC-15388 - Update to Yarn 4

axeleriksson147:
This PR is up for discussion but I was having major problems linking Yarn 1 to Yarn 4 that we use in SFE-RTC. Yarn 1 entered maintenance mode 5 years ago.


.yarnrc.yml
```diff
@@ -0,0 +1,4 @@
+compressionLevel: mixed
+enableGlobalCache: false
+nodeLinker: node-modules
+npmRegistryServer: "https://registry.yarnpkg.com"
```
axeleriksson147, line 4: This actually works and doesn't default to the Symphony private registry if configured.

.gitignore
```diff
@@ -12,3 +12,7 @@ dist
 symphony-uitoolkit-*.tgz
 icons.stories.js
 package-lock.json
+
+yarn-error.log
```
axeleriksson147, line 16: When using nodeLinker `node-modules` these yarn files can be ignored.

yarn.lock
axeleriksson147, line 1: The yarn.lock structure is completely different.

PR #722, opened 2025-05-19
Make use of dropdown input label to improve clear button accessibility

kgning1:
## Description

This change uses dropdown input label when available to improve clear button description in voice over

## JIRA ticket

[CES-22163](https://perzoinc.atlassian.net/browse/CES-22163)

image

[CES-22163]: https://perzoinc.atlassian.net/browse/CES-22163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@symphony-ui/uitoolkit",
-  "version": "4.9.6",
+  "version": "4.9.7",
   "private": true,
   "description": "React components for the Symphony UI design system",
   "license": "ISC",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@symphony-ui/uitoolkit-components",
-  "version": "4.9.6",
+  "version": "4.9.7",
   "description": "React components for the Symphony UI design system",
   "publishConfig": {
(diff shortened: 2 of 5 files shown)


antoinerollindev (approved):

benoit-sy (commented):
As a matter of fact, this kind of feature as usually not made accessible to screen reader. most accessible libraries don't have this behavior.

the clear button is more for mouse users that will click on it to clear. keyboard users are going to use the keyboard itself to clear the field. these buttons in most libraries are not even focusable with keyboard.

have a look at https://react-spectrum.adobe.com/react-aria/SearchField.html for example (which is one of the most complete accessible React widget library)

Making these buttons as part of the tab navigation is more a drawback to me and I think we should not make this change.

kgning1:
Hello @benoit-sy thanks for the review unfortunately this PR is meant to respond to one of our client request for accessibility . Do you have a suggestion to maybe make this behavior optional or have a similar result ?

benoit-sy:
if it's a client's request, then I am afraid we should do it. that's unfortunate.

PR #728, opened 2025-06-23
UPL-151: `DateTime` component

antoinerollindev:
[UPL-151](https://perzoinc.atlassian.net/browse/UPL-151): `DateTime` component

New `DateTime` component allowing to display a date or a time (or both) according to the provided settings. By default, renders date and time using the browser locale formatting.

Properties:
- `date`: the date information to display (ISO string or Date)
- `format`: one of:
  - `DATE_TIME` (default): renders the date and time (i.e. '12/31/2024 10:29 AM')
  - `DATE`: renders the date (i.e. '12/31/2024')
  - `TIME`: renders the time (i.e. '10:29 AM')
  - `TIME_SECS`: renders the time with seconds (i.e. '10:29:47 AM')
- `locale`: the locale to use for date and time formatting (i.e. 'fr-FR') - default is the browser locale
- `show12HourTime`: whether to display the time in 12-hour format (i.e. '05:29 PM')
- `dateFormat`: the format to use for the date (i.e. 'yyyy/MM/dd') - overrides the locale formatting
- `timeFormat`: the format to use for the time (i.e. 'hh:mm:ss a') - overrides the locale formatting, the `show12HourTime` prop and may affect the display of time seconds

-------

https://github.com/user-attachments/assets/3472e2a8-f2b9-495f-b73f-268db39bfaae

[UPL-151]: https://perzoinc.atlassian.net/browse/UPL-151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/spec/components/date-time/DateTime.spec.tsx
+++ b/packages/components/spec/components/date-time/DateTime.spec.tsx
@@ -1,0 +1,121 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { DateTime } from '../../../src/components';
+import { DateTimeFormat } from '../../../src/components/date-time/interfaces';
+
+describe('DateTime', () => {
+  const testDate = new Date(2024, 11, 30, 17, 30, 15);
+  const locale = 'en-US';
+
+  const defaultProps = {
+    date: testDate,
+    locale,
+  };
+
+  describe('date', () => {
+    it('should handle Date objects', () => {
+      const { container } = render(<DateTime {...defaultProps} />);
+      expect(container).toHaveTextContent('12/30/2024 05:30 PM');
+    });
+
+    it('should handle ISO string dates', () => {
+      const dateISOString = testDate.toISOString();
+      const { container } = render(
+        <DateTime {...defaultProps} date={dateISOString} />
+      );
+      expect(container).toHaveTextContent('12/30/2024 05:30 PM');
+    });
+  });
+
+  describe('format', () => {
+    it('should display date and time by default', () => {
+      const { container } = render(<DateTime {...defaultProps} />);
+      expect(container).toHaveTextContent('12/30/2024 05:30 PM');
+    });
+
+    it('should display date only', () => {
+      const { container } = render(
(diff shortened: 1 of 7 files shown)


antoinerollindev:
> Just took a quick glance but looks good overall.
> 
> https://blog.date-fns.org/v40-with-time-zone-support/ - The `DateTime` component is pretty much a wrapper for the `date-fns` package. Does the newer version provide anything where we don't need to write our own interfaces and utils and we could extend theirs instead?
> 
> I guess when https://developer.mozilla.org/en-US/blog/javascript-temporal-is-coming/ starts getting supported in browsers there won't be really any need for the component as it seems like this has wide support for setting the correct time format. But that's not for a long time given the versions we have to support, but just interesting new web API.

Yes the `DateTime` component is mainly using `date-fns` for applying a format (like 'mm/dd/yyyy') when it's provided.
However, date and time format props are optional, and we want to fallback to the browser locale. The `date-fns` locale management is quiet heavy as you have to do a mapping between locale strings and date-fns locale objects. This is why I prefer using the `Date.toLocaleString` function which is handling the default locale and applies the proper formatting.

The implicit goal here is to render this component from Nexus when a message contains a `<time>` element (new presentationML element). Also, having such a component is interesting if you want to avoid doing all this date/time display logic on your own.

packages/components/spec/components/date-time/DateTime.spec.tsx
```diff
@@ -0,0 +1,121 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { DateTime } from '../../../src/components';
+import { DateTimeFormat } from '../../../src/components/date-time/interfaces';
+
+describe('DateTime', () => {
+  const testDate = new Date(2024, 11, 30, 17, 30, 15);
+  const testISOString = testDate.toISOString();
```
axeleriksson147, line 8: Nitpicking. Feel free to edit or not, no big deal.

But this should be declared in the test that uses it. No point in having it declared for all test files.

PR #729, opened 2025-06-24
RTC-15509 - Not trigger onClose for DropdownMenu if event comes from a Modal

axeleriksson147:
https://perzoinc.atlassian.net/browse/CES-22535
https://perzoinc.atlassian.net/browse/RTC-15509

- An item in a `DropdownMenu` opens a `Modal` component. When closing that `Modal`, it would also close the `DropdownMenu`. If we check the composedPath for `tk-dialog-backdrop` we know the mouse event or keyboard event stems from the `Modal`.

Update: before merging this one I would like to know where does this requirement come from. Is this truly the expected behaviour?


axeleriksson147:
This PR can potentially be closed if everyone agrees that this is not intended behaviour.

PR #743, opened 2025-08-05
Adding optional aria label property to dropdown input

kgning1:
## Description
Adding optional aria label property to dropdown input in case we want to customize text read by accessibility tools without displaying it.

## JIRA ticket

[CES-23245](https://perzoinc.atlassian.net/browse/CES-23245)

[CES-23245]: https://perzoinc.atlassian.net/browse/CES-23245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/dropdown/CustomRender.tsx
+++ b/packages/components/src/components/dropdown/CustomRender.tsx
@@ -62,7 +62,7 @@ export const DefaultOptionRenderer = (props: any) => {
 // See https://github.com/JedWatson/react-select/discussions/4302
 export const Input = (props: any) => {
   const inputAlwaysDisplayed = props?.selectProps?.inputAlwaysDisplayed;
-  const ariaLabel = props?.selectProps?.label? props.selectProps.label : props?.selectProps?.ariaLabel;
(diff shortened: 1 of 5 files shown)


packages/components/src/components/dropdown/CustomRender.tsx
```diff
@@ -62,6 +62,7 @@ export const DefaultOptionRenderer = (props: any) => {
 // See https://github.com/JedWatson/react-select/discussions/4302
 export const Input = (props: any) => {
   const inputAlwaysDisplayed = props?.selectProps?.inputAlwaysDisplayed;
+  const ariaLabel = props?.selectProps?.label? props.selectProps.label : props?.selectProps?.ariaLabel;
```
aurelie-forzani, line 65: I would to the opposite, use ariaLabel and if ariaLabel is not defined, then fallback to label

antoinerollindev (approved):

